PR #826, opened 2026-04-14
fix(statusline): restore OSC 8 clickable hyperlinks for Dashboard/Lessons

IgorGanapolsky:
Cherry-pick of 0761da3c. OSC 8 links were lost — Dashboard/Lessons/thumbs rendered as plain text instead of clickable hyperlinks in WezTerm/iTerm2.

### DIFF
--- a/.changeset/osc8-links.md
+++ b/.changeset/osc8-links.md
@@ -1,0 +1,5 @@
+---
+"thumbgate": patch
+---
+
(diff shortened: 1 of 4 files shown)


IgorGanapolsky:
Closing as superseded by #827, which restored OSC 8 clickable Dashboard/Lessons labels without leaking localhost URLs into the statusline. This branch is also conflicting with main, so keeping it open would create duplicate stale PR noise.

PR #1142, opened 2026-04-22
fix: allow pending changesets in publish guard

IgorGanapolsky:
## Summary
- let the publish no-op guard accept pending valid .changeset entries on main
- keep blocking true silent no-op publishes when shipped content drifts without release metadata
- pin the workflow contract with deployment tests

## Verification
- node --test tests/deployment.test.js
- npm run test:deployment


IgorGanapolsky:
Closing as duplicate: origin/main already contains fix(ci): allow changeset-audited publish no-op via #1138, and the current main build lane is green.

PR #1180, opened 2026-04-22
fix(api): lazy-load billing private boundary

IgorGanapolsky:
## Summary
- move billing routes behind the private API loader instead of direct public-repo imports
- preserve the hosted/local billing flows while returning the standard private-core contract when billing is unavailable
- add API coverage for the new billing private-core boundary

## Verification
- `node --test --test-name-pattern='billing session endpoint returns provisioned local checkout details|billing checkout supports annual Pro and Team seat selection|billing checkout supports CORS preflight|billing session endpoint rejects missing session ids|billing provision requires static admin key and rejects billing keys|billing summary returns admin-only operational proxy|billing summary applies today window query params for admin users|dashboard applies analytics window query params with live billing truth|dashboard render-spec endpoint returns constrained hosted views|billing summary includes Stripe-reconciled revenue when live processor events are available|billing summary rejects billing keys|billing summary rejects invalid analytics window queries|funnel analytics returns counts and conversion rates' tests/api-server.test.js`
- `npm run test:deployment`
- `npm pack --dry-run --json --ignore-scripts | jq '.[0] | {fileCount: (.files|length), unpackedSize}'`
- `npm run changeset:check`


IgorGanapolsky:
Stale lazy-load fix: superseded by recent billing boundaries

PR #1539, opened 2026-05-02
feat(gtm): preserve business-system marketplace surfaces

IgorGanapolsky:
## Summary
- keep the business-system workflow approvals segment visible in marketplace copy even when it falls outside the raw top-three signal cutoff
- add a regression test that pins both the business-system and self-serve variants through that cutoff case
- refresh the generated GTM revenue-loop, queue, handoff, send-now, and marketplace assets from the current evidence-backed report

## Verification
- node --test tests/gtm-revenue-loop.test.js


IgorGanapolsky:
Closing as stale DIRTY (≥5 days, conflicting with current main). Branch remains on remote for archival.

PR #1770, opened 2026-05-06
fix(gtm): make revenue sends fail loudly

IgorGanapolsky:
## Summary\n- fail revenue email dispatches when Resend/provider refuses the send\n- add bounded safe auto-approval for Bluesky reply drafts\n- expose a confirmed autonomous Bluesky publish path in the reply monitor workflow\n\n## Tests\n- node --test tests/social-reply-monitor-bluesky.test.js tests/revenue-email-dispatch.test.js tests/workflow-runs.test.js\n- git diff --check

### DIFF
--- a/.changeset/autonomous-bluesky-replies.md
+++ b/.changeset/autonomous-bluesky-replies.md
@@ -1,0 +1,5 @@
+---
+"thumbgate": patch
+---
+
+Add a guarded autonomous Bluesky reply publishing path and fail revenue email dispatches when the mail provider refuses the send.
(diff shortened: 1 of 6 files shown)


IgorGanapolsky:
Closing as stale (DIRTY 5-8 days, ~60 main commits since opened). The branch remains on the remote for archival; to revive, rebase onto current main and reopen. Replacement scope: revenue sends fail loudly may need fresh implementation against current architecture rather than conflict-resolution.

PR #1772, opened 2026-05-06
feat(revenue): add checkout recovery downsells

IgorGanapolsky:
## Summary\n- add , , and  paid recovery options to the checkout intent interstitial\n- add the same low-friction paid recovery links to the checkout-cancel page\n- preserve attribution on checkout interstitial links for campaign measurement\n\n## Tests\n- npm run test:checkout-bot-guard\n- node --test tests/api-server.test.js\n- pre-commit guard\n- pre-push guard


IgorGanapolsky:
Replaced by clean PR #1773 based on current main with only the checkout recovery downsell change.

PR #2057, opened 2026-05-14
feat(revenue): surface $19 Quick Read upsell at thumbs-down pain point

IgorGanapolsky:
## Summary

ThumbGate has 760+ weekly npm downloads. The /pro page already wires the $19 Quick Read Stripe Payment Link. But free users live in the CLI, not the marketing page — they never see the offer at the moment of intent.

This puts the offer where the pain lives: right after `capture-feedback --feedback=down`, if the operator has logged 3+ thumbs-downs in the last 24h and isn't Pro-licensed.

## What changes

- **New module** `scripts/upsell-quickread.js` — `maybePrintUpsell()` checks (a) recent thumbs-down count in last 24h, (b) `isProLicensed()`, (c) cooldown timestamp at `~/.thumbgate/last-upsell.json`, (d) `THUMBGATE_NO_UPSELL` env. Prints a single-screen offer to **stderr only** when all four say go.
- **Wired into** `.claude/scripts/feedback/capture-feedback.js` after a successful thumbs-down capture. Wrapped in try/catch so capture-feedback never breaks on upsell errors.
- **12 tests** covering window filtering, malformed-log tolerance, env suppression, Pro suppression, threshold, cooldown enter/exit, missing log, message formatting.

## Why this should make money (honest assessment)

- The /pro page CTA converts at near-zero because most npm users never load it. This puts the offer in the place they live (their terminal) at the moment they've just demonstrated a recurring failure pattern.
- Even a 0.5% click-through on 760 weekly downloads is ~4 paying customers/week at $19 = $300+/mo within 30 days *if* the wedge works.
- If it doesn't work after 30 days, we have data and can change the threshold, copy, or feature gate.

## Guardrails

- Stderr-only (script consumers parsing JSON output unaffected)
- Suppressed for Pro users via `isProLicensed()`
- Rate-limited 1x per 24h via timestamp file
- Opt-out: `export THUMBGATE_NO_UPSELL=1`

## Test plan

- [x] 12 unit tests pass (`npm run test:upsell-quickread`)
- [x] End-to-end smoke: synthetic feedback log → upsell fires with correct count and link
- [x] Wired into `test:high-roi` suite chain
- [x] package.json `files` whitelist updated for npm bundle
- [x] Changeset added (minor bump)


IgorGanapolsky:
Holding this PR. The branch is mergeable but it's a downstream conversion optimization — meaningless without top-of-funnel pipeline. Will revisit after 10 real outbound conversations have happened. See thread for context.

PR #2176, opened 2026-05-19
feat: Bayesian A/B testing harness on top of Beta-binomial conversion stats

IgorGanapolsky:
## What

`scripts/ab-experiment-framework.js` — the experiment harness on top of the Beta-binomial posterior math that already ships in `scripts/conversion-rate-stats.js`.

## Why

ThumbGate already has the conversion-rate math (Beta-binomial posteriors with credible intervals). What was missing was the *experiment harness*: variant assignment, decision rule, sample-size guard, stopping criteria. Without it, every pricing/copy change is a coin flip and we can't claim a result is real.

This is the first of the 5 "adequate analytics" items previously listed as missing.

## How it decides

1. **Variant assignment** — deterministic SHA-256 hash of `experimentKey:visitorId`. Same visitor → same variant for the experiment's lifetime. No server-side sticky session.
2. **Decision rule** — Bayesian. We compute `P(rate_B > rate_A)` by Monte Carlo sampling from the Beta posteriors. No p-values, no fixed sample size, no peeking penalty.
3. **Practical-significance guard** — require posterior probability of a *meaningful* lift (default 5% relative) to clear threshold (default 0.95). Statistical significance ≠ business significance.
4. **Loss-bound stopping** — if neither arm wins but expected loss of picking the wrong arm falls below ε (default 0.05% absolute), stop. The experiment is decided to be inconclusive because the choice no longer matters.

Verdicts: `ship_a` | `ship_b` | `inconclusive_stop` | `continue`.

## Tests

11 tests, all green:

- Variant assignment determinism + cross-experiment independence (1000 visitors, ~50% disagreement confirms hash independence)
- Weight calibration (5k samples, 80/20 split, ~20% ±3pp tolerance)
- `decide` ships B when B has 10% vs A's 5% rate
- `decide` ships A when A is clearly better
- `decide` stops as inconclusive at 100k×100k tied trials
- `decide` says continue at low sample size with no clear winner
- Markdown renderer + CLI parsing

## Out of scope

Page-rendering integration (cookie set, variant-aware HTML) is the next PR. This lands the foundation so that change is a thin server-side wire-up.


IgorGanapolsky:
Closing stale DIRTY PR — merge conflicts with main. Will re-branch if needed.

PR #2189, opened 2026-05-19
fix: gate /success conversion metric on session_id presence + bot check

IgorGanapolsky:
## The bug

Today's revenue audit (2026-05-19 17:11 UTC):

| Window | checkoutSuccessPageViews | checkoutPaidConfirmations |
|---|---:|---:|
| Today | 1 | 0 |
| 30 days | **6** | **0** |

`/success` is a publicly-hittable URL. Direct nav, bot crawls, monitoring probes, broken back-buttons, copy-pasted shared links — all of them landed on `/success` and the server unconditionally emitted `checkout_success_page_view`. The audit aggregates that into the conversion funnel. So the metric was indistinguishable from noise: we knew the page loaded, not whether anyone paid.

In a world where real paid conversions are 0, the noise eats the signal entirely.

## Fix

`src/api/server.js` `/success` handler now:

```js
const sessionIdParam = parsed?.searchParams?.get('session_id') || '';
const successBotClass = classifyRequester(req.headers);
const isCountedSuccess = sessionIdParam.startsWith('cs_') && !successBotClass.isBot;
const successEventType = isCountedSuccess
  ? 'checkout_success_page_view'
  : 'checkout_success_page_view_unverified';
```

- **`cs_` prefix** covers both `cs_live_*` (production) and `cs_test_*` (test mode). It is exactly the prefix Stripe uses on its post-payment redirect URL.
- **`!isBot`** keeps Googlebot/ClaudeBot/Slackbot/etc. out of the conversion count even if they happen to hit a URL with a `cs_`-prefixed query.
- Unverified hits still emit telemetry under a parallel event name so we keep observability for raw traffic — they just don't inflate the conversion metric.

## Tests

`tests/checkout-success-noise-gate.test.js`, 5 cases (4 unverified, 1 verified, 1 test-mode parity):

```
✔ does NOT emit checkout_success_page_view when session_id is absent
✔ does NOT emit canonical event when session_id has a non-Stripe prefix
✔ does NOT emit canonical event for a bot UA even with a real-shaped session_id
✔ DOES emit canonical checkout_success_page_view for human UA + cs_-prefixed session_id
✔ accepts cs_test_ session_id prefix (Stripe test mode) the same as cs_live_
```

Wired into the `npm test` chain via `test:checkout-success-noise-gate`. `test-suite-parity` 5/5 pass.

## Adjacent suites verified (origin/main `bf964cff` base)

- `test:checkout-bot-guard` 13/13
- `test:checkout-pro-confirmation-gate` 3/3
- `test:funnel-invariants` 15/15
- `test:revenue-observability-doctor` 5/5

## What this does NOT do (deliberate scope cap)

It does not verify the session against Stripe — a `stripe.checkout.sessions.retrieve(session_id)` probe would be stronger but adds Stripe API latency to every pageview. This patch is the cheap, deterministic noise gate; the verifying-call upgrade can land as a follow-up once paid traffic justifies the call cost.

## Expected outcome

After deploy + 24 hours of audits, `checkoutSuccessPageViews` on the 30-day window should:
- DROP toward 0 if the prior 6 were all noise (most likely)
- Or hold steady but with the `*_unverified` event count rising proportionally, meaning we previously had a few real paid users we weren't surfacing

Either result is a strictly better signal than what we have now.


IgorGanapolsky:
Closing stale DIRTY PR — merge conflicts with main. Will re-branch if needed.

PR #2242, opened 2026-05-20
fix(dashboard): wire stat-card filters + Playwright E2E + CI

IgorGanapolsky:
## Summary

- **Bug fix**: Operations Dashboard stat cards (`Positive` / `Negative` / `Total Feedback`) navigated to `/lessons` but the lessons page never read the `?signal` query param — every card landed on the Rules tab unfiltered. CEO reported via screenshot: clicking POSITIVE did not show positives.
- **Test infrastructure**: Added Playwright 1.x with chromium, `page.route()` API mocking, fully-parallel local execution, 2-shard matrix in CI.
- **CI**: New `.github/workflows/e2e.yml` runs on pushes / PRs that touch `public/`, `src/api/`, `tests/e2e/`, `playwright.config.js`, or `package.json`. Sharded matrix with blob reporter + dependent merge-reports job. Browser cache keyed on `@playwright/test` version.

## Root cause

`public/dashboard.html:251-252` emitted `?signal=positive|negative`. The rest of the codebase uses canonical `up|down` vocabulary. `public/lessons.html` never parsed the query param at all — `loadLive().then(...)` went straight to `highlightCard(0)` (Rules tab). Same root cause hit Total Feedback and Negative cards.

## Minimal fix

- `public/dashboard.html` — switch hrefs to `?signal=up|down|all` (canonical vocab).
- `public/lessons.html` — at bootstrap, read `?signal`, accept canonical + legacy aliases, call `switchTab('timeline') + filterTimeline(mapped)`.

## Rebase note (history was rewritten)

Earlier revisions of this branch bundled an unrelated `fix(sentinel)` commit (`7f09dce7`) pushed by a parallel agent session, plus follow-up version-sync and codex-pack regen commits that became obsolete once main shipped v1.21.2 via #2241. The branch has been **force-rebased onto current main** (1.21.2) and reduced to two commits:

- `8378c588` — dashboard fix + Playwright suite + e2e.yml workflow
- `f29148b4` — changeset entry

The original sentinel commit is preserved on tag [`wip/sentinel-from-parallel-session-2026-05-20`](https://github.com/IgorGanapolsky/ThumbGate/releases/tag/wip/sentinel-from-parallel-session-2026-05-20) so the parallel session can re-PR it cleanly.

## Verification

- [x] Playwright suite green (5/5 specs locally on rebased state).
- [x] All 128 existing tests across the 13 dashboard/lessons-touching test files pass locally.
[...]

PR #2257, opened 2026-05-20
fix(build-metadata): file-baked SHA wins over runtime env (anti-drift)

IgorGanapolsky:
**Why**

Prod \`/health\` has been lying for hours: returns \`version=1.21.2\` but \`buildSha=92f8e4b1...\` — a commit from days before the v1.21.2 release. Verified the actual deployed code IS new (the dashboard click fix from #2242 is live), but the SHA telemetry is stale.

**Root cause**

\`resolveBuildMetadata()\` in \`scripts/build-metadata.js\` reads env vars first, JSON file second. The Railway service has a stale \`THUMBGATE_BUILD_SHA=92f8e4b1\` env var from a previous deploy, and the \`Set Railway environment variables\` step in \`.github/workflows/deploy-railway.yml\` is gated by \`RAILWAY_SYNC_VARIABLES=false\` (default) — so the env var never gets refreshed, shadowing the freshly-stamped JSON file forever.

Secondary bug: condition \`envBuildSha || envGeneratedAt\` short-circuits to \`{ buildSha: null }\` if only a stray \`GENERATED_AT\` env is set, losing both signals.

**Fix**

Invert precedence: the immutable JSON file baked into the Docker image at build time (which **always** matches the deployed code) wins over mutable runtime env vars. Env vars now fill in only when the file has no SHA, AND the env branch requires an explicit SHA (no more null-return when only GENERATED_AT is set).

**Side benefit — unblocks \`Verify deployment health\`**

That step compares \`LIVE_SHA\` to \`\$GITHUB_SHA\` (line 313 of deploy-railway.yml). With the env-wins bug, it would compare \`92f8e4b1\` against the actual merge commit and fail — likely contributing to today's repeated deploy failures.

**Tests**

- Existing roundtrip test still passes.
- New tests: file-wins (anti-drift), env-fallback when file empty, regression for stray-generatedAt short-circuit.
- 7/7 pass locally.

**Operator follow-up (no code change required)**

Once this merges + deploys, you can safely delete the persistent \`THUMBGATE_BUILD_SHA\` and \`THUMBGATE_BUILD_GENERATED_AT\` env vars on the Railway service via the dashboard. File precedence makes them moot. (Optional cleanup, not required for the fix to work.)

:robot_face: Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/build-metadata-file-wins-precedence.md
+++ b/.changeset/build-metadata-file-wins-precedence.md
@@ -1,0 +1,13 @@
+---
+"thumbgate": patch
+---
+
+Fix `/health` reporting the wrong `buildSha` when a stale `THUMBGATE_BUILD_SHA` env var lingers on the runtime host. Invert precedence in `scripts/build-metadata.js`: the immutable JSON file baked into the Docker image at build time (which always matches the deployed code) now wins over mutable runtime env vars. Env vars fill in only when the file has no SHA.
+
+Also tightens the env-branch condition: previously a stray `THUMBGATE_BUILD_GENERATED_AT` with no SHA would short-circuit to `{ buildSha: null }`, losing both signals. Now the env branch requires an explicit SHA before being trusted.
+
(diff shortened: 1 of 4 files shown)


IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `911aee76d0c5747ab58a096d5a57670abd56303c`
**Production buildSha:** `911aee76d0c5747ab58a096d5a57670abd56303c`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","version":"1.21.2","buildSha":"911aee76d0c5747ab58a096d5a57670abd56303c","uptime":21.425480229,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

PR #2268, opened 2026-05-20
fix(lessons): stat-card clicks scroll content into view + 17 E2E tests covering every clickable surface

IgorGanapolsky:
## Why

CEO clicked the four stat tiles on `/lessons` ("Active Rules", "Critical", "Actions Blocked", "Approval Trend") and said *"nothing happens"*. The handlers were calling `switchTab()` correctly — but the tab content was below the fold, the page never scrolled to it, and the user-visible state didn't change.

Compounding issue: my Playwright suite from PR #2242 only covered the `/dashboard` stat cards. The `/lessons` tiles, tab headers, and filter buttons had **zero** E2E coverage. CEO called this out directly: *"your playwright suite needs 100% e2e verification!!!!!"*

### Fix
- `public/lessons.html` `switchTab()`: after flipping the active class, call `content.scrollIntoView({ behavior: 'smooth', block: 'start' })`. Click now has visible effect.

### Coverage
- `tests/e2e/lessons-page-clickability.spec.js` (new) — **17 tests** covering every deterministic clickable surface on `/lessons`:
  - 4 stat tiles (each asserts `getBoundingClientRect` proves scrollIntoView landed)
  - 3 tab headers (Active Rules, Feedback Timeline, Insights)
  - 4 rules filter buttons (All / Critical / Warning / Info)
  - 3 timeline filter buttons (All / 👍 Positive / 👎 Negative)
  - 2 nav anchors (Dashboard, Landing Page)
  - 1 render check on the four stat values
- 17/17 pass locally.

## Honest acknowledgment

PR #2242 claimed "thorough E2E testing" but the suite only exercised the **specific bug being fixed**, not the broader **bug class** (every clickable stat tile in the ThumbGate UI). The CEO's mental model is *"stat cards in ThumbGate should be clickable and do something visible"* — and that was only half-covered. This PR closes the gap for `/lessons`. A follow-up audit should also widen coverage on `/`, `/agent-manager`, `/codex-enterprise`, `/pricing` if those surfaces have similar tile patterns.

Lesson logged to feedback DB: when writing E2E for a clickable surface, enumerate ALL clickable surfaces in the same UI category and assert each has a user-visible effect (URL change, scroll change, content change) — never just "the handler fires."

## Files

| File | What |
|---|---|
| `public/lessons.html` | switchTab adds scrollIntoView |
| `tests/e2e/lessons-page-clickability.spec.js` | 17 new Playwright tests |
| `package.json` | added `test:lessons-page-clickability` script |
[...]

PR #2276, opened 2026-05-21
fix(dashboard): Active Gates card click activates tab + scrolls into view

IgorGanapolsky:
## What

CEO reported clicking the **Active Gates** stat card on the localhost dashboard does nothing. Confirmed bug in two places inside `switchTab()` (`public/dashboard.html`).

## Root cause

```js
// Before
var tabEl = document.querySelector('[onclick*="' + name + '"]');
```

For `name = 'gates'`, this selector also matches the stat-card on line 253 (`onclick="selectCard(this,'gates');return false;"`). The card appears **before** the tab header in DOM order, so `querySelector` returns the card. The card gets `.active` (no visible effect, it's not styled as a tab), and the actual tab header stays dormant.

The tab CONTENT panel (`#tab-gates`) did get `.active` via `getElementById`, but it sits below the fold — without `scrollIntoView`, the click felt like a no-op.

## Fix

1. Scope the header lookup to `.tab[onclick*="<name>"]` so stat-cards can't match.
2. Add `contentEl.scrollIntoView({ behavior: 'smooth', block: 'start' })` so the freshly-activated panel lands in the viewport. Same class of fix as PR #2268 (/lessons tile fix).

## Tests

Two new tests in `tests/e2e/dashboard-stat-cards.spec.js` pin the regression — both fail against `main`, both pass with this fix. All 7 dashboard-stat-card tests pass locally.

### DIFF
--- a/.changeset/fix-active-gates-card-click.md
+++ b/.changeset/fix-active-gates-card-click.md
@@ -1,0 +1,20 @@
+---
+"thumbgate": patch
+---
+
+fix(dashboard): Active Gates stat-card click now activates the Gates tab + scrolls it into view
+
+Clicking the Active Gates card on /dashboard previously appeared to do nothing.
+Two bugs in switchTab():
+
+1. The selector `document.querySelector('[onclick*="<name>"]')` matched the
+   stat-card (first in DOM order) instead of the tab header for that name,
+   so the tab header never lit up as active.
+2. The tab content panel did get .active, but it sat below the fold; with
+   no scrollIntoView, the user perceived "nothing happened" because the
+   newly-visible content was off-screen.
+
+Fix: scope the header selector to `.tab`, and call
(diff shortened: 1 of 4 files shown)


IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `5cda284d7dfab2aa835896a73f20b94020145a9c`
**Production buildSha:** `5cda284d7dfab2aa835896a73f20b94020145a9c`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","version":"1.21.2","buildSha":"5cda284d7dfab2aa835896a73f20b94020145a9c","uptime":113.643753319,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

PR #2288, opened 2026-05-21
feat(cli): seamless auto-update — every CLI invocation refreshes thumbgate@latest in background

IgorGanapolsky:
Closes the gap left by the Volta-style hook shim. Hook invocations already auto-update via `scripts/install-shim.js`; this does the same for direct CLI calls (`thumbgate cost`, `thumbgate stats`, etc.) which previously ran whatever was last globally installed.

## Behavior
- Non-blocking — zero added latency on the user's command
- Throttled to ≤1 install attempt per 24h (marker at `~/.thumbgate/.last-update-check`)
- Permission-safe — silent failure if npm install fails (no sudo prompts)
- Opt-out via `THUMBGATE_NO_AUTO_UPDATE=1`
- Auto-skipped in CI envs and `NODE_ENV=test`

## Files
- `scripts/auto-update-cli.js` (~100 LOC) — the throttled fire-and-forget module
- `bin/cli.js` — wired at the top via try/require so a failure never blocks the CLI
- `tests/auto-update-cli.test.js` — 12 tests, all pass locally
- `package.json` — adds `test:auto-update-cli` + appends to master test chain


IgorGanapolsky:
Rebased this branch against current `main` and pushed cleanly. CI re-ran on the rebased commit and `test` failed again, same shape as the dependabot PRs earlier today (5 of them closed for the same pattern: `test` job fails in CI in ~2-5 min, full `npm test` chain passes locally on current main).

I tried to reproduce `node --test tests/auto-update-cli.test.js` locally on this branch but the container SIGTERM'd the process — looks like the auto-update test calls `spawn('npm', ['install', '-g', '--silent', '--no-audit', '--no-fund', 'thumbgate@latest'], { detached: true })` which actually attempts a real global npm install. In a sandboxed CI runner that's the most likely failure mode: the spawned npm child either errors on permissions or hangs, and the test framework gives up.

Gitar already flagged exactly this surface as a 💡 finding (scripts/auto-update-cli.js:74-87): "Full process.env forwarded to detached background npm process" + suggested fix is a minimal env allowlist. That fix probably also wouldn't be enough to make tests pass — the bigger issue is that the test runs the actual install instead of stubbing the `spawn` call.

**Two recommended paths for you, take your pick:**

1. **Stub `spawn` in the tests.** Replace `tests/auto-update-cli.test.js` calls to `maybeTriggerBackgroundUpdate({ ... })` with explicit injection of a no-op spawner. The implementation already takes options; tests should pass `{ spawnImpl: () => ({ unref(){}, on(){} }) }` or similar so the real npm never runs.
2. **Skip the spawn-touching tests in CI** via `process.env.CI ? skip : run`. Cheap but masks the real issue.

I won't autonomously push (2) — it hides the bug — and (1) is a real test refactor that you should sign off on. Branch is rebased and pushed at SHA `e891b28`, so when you fix the test path it's ready to land.

---
_Generated by [Claude Code](https://claude.ai/code)_

IgorGanapolsky:
Closing as **superseded by `main`** (verified by merging current main — branch was 106 behind — and measuring net delta).

The auto-update feature already ships in main, more robustly: `scripts/published-cli.js` `publishedCliShellCommand(..., {preferInstalled:false})` is wired via `codexAutoUpdateCliEntry` in `scripts/mcp-config.js` (installs `thumbgate@latest` on every CLI/MCP invocation) plus the `THUMBGATE_SHIM_AUTO_UPDATE` shim in `scripts/install-shim.js`. Main's approach uses a private runtime prefix (no global-npm/sudo risk) and is publication-gated.

This branch's only net-new files are `scripts/auto-update-cli.js` (a detached `npm install -g thumbgate@latest`), its test, and a changeset — and the `bin/cli.js` wiring that was supposed to call it lost the merge to main, leaving the script **orphaned dead code** (referenced nowhere, absent from `package.json` files[], no `test:` script). Reopen if I've missed unique value.

PR #2554, opened 2026-06-06
fix(statusline): stop double-counting cached aggregates (1152→727 bug)

IgorGanapolsky:
## Symptom

CEO's statusline screenshot showed `1152👍 747👎 ↘`. The real cross-store aggregate is `727👍 / 600👎`. Investigation traced the inflated value to `scripts/statusline-cache-read.js`.

Reproducible on the same host:

```
$ node scripts/statusline-cache-read.js | jq '{thumbs_up, thumbs_down, sources_count}'
{
  "thumbs_up": "1152",
  "thumbs_down": "747",
  "sources_count": 4    ← summed 4 caches where the first IS already the aggregate
}

$ node -e "const m=require('./scripts/feedback-aggregate'); const s=m.computeAggregateFeedbackStats({cwd:process.cwd()}); console.log(s.totalPositive, s.totalNegative)"
727 600                ← the true cross-store total
```

## Root cause

`feedback-aggregate.js` and `hook-thumbgate-cache-updater.js` write the cross-store aggregate to `~/.thumbgate/statusline_cache.json`. That file already contains the correct total. The per-folder caches (`<project>/.thumbgate/statusline_cache.json`) are subsets of it.

`statusline-cache-read.js` was summing `thumbs_up`/`thumbs_down` across all of them, double- and triple-counting every event.

## Fix

`statusline-cache-read.js` now resolves the highest-priority existing cache from the candidate list (which already puts the canonical aggregate first when aggregation is enabled) and returns its content unchanged. No summing — the upstream aggregator already did that work.

## Tests

- Rewritten `tests/statusline-cache-aggregate.test.js` (4 tests) to **pin non-summing behavior**: when both the canonical aggregate and per-folder subsets exist, the resolved value MUST equal the aggregate (727), not the sum (727+subsets).
- The original tests asserted the sum and would have caught nothing. The new tests would have caught the 1152 bug on day one.

## Out of scope (filed separately in the body of this PR)

The launchd dashboard agent runs from the **npm-installed thumbgate v1.27.6**, which predates `feedback-aggregate.js` entirely. It overwrites `~/.thumbgate/statusline_cache.json` with old per-folder values regardless of source-repo state. This PR doesn't change that — **a new npm release is required for end users to see the corrected number in their statusline.** After this lands, suggest `npm version patch && npm publish`.

### DIFF
--- a/.changeset/statusline-cache-no-double-count.md
+++ b/.changeset/statusline-cache-no-double-count.md
@@ -1,0 +1,5 @@
+---
+"thumbgate": patch
+---
+
+Fix statusline display showing inflated thumbs counts (e.g. 1152↑/747↓ when the true cross-store total was 727↑/600↓). `scripts/statusline-cache-read.js` was summing `thumbs_up`/`thumbs_down` across every per-folder `statusline_cache.json`, but the canonical global aggregate at `~/.thumbgate/statusline_cache.json` is itself written as the cross-store sum (by `feedback-aggregate.js`). Summing the aggregate plus the per-folder caches counted every event twice. The helper now resolves the highest-priority existing cache and returns its content unchanged.
--- a/scripts/statusline-cache-read.js
+++ b/scripts/statusline-cache-read.js
@@ -1,62 +1,24 @@
 #!/usr/bin/env node
 'use strict';
 
+// Resolve the statusline cache to read for display.
+//
+// PRIOR BUG: the earlier version of this script "aggregated" by summing the
+// thumbs_up/down fields across every per-folder statusline_cache.json. That
+// double-counted, because the global aggregate cache at
+// ~/.thumbgate/statusline_cache.json is ITSELF already the cross-store sum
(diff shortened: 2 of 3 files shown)


IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `dcc8d66ed9e5004e39ea60f91b87bab147d886ea`
**Production buildSha:** `dcc8d66ed9e5004e39ea60f91b87bab147d886ea`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","degraded":false,"version":"1.27.6","buildSha":"dcc8d66ed9e5004e39ea60f91b87bab147d886ea","uptime":18.988561469,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

PR #2565, opened 2026-06-06
feat: agent work-claim lock to prevent concurrent-agent double-work

IgorGanapolsky:
## What

Adds an **agent work-claim lock**: a unit of work can be CLAIMED by one agent before it is done, so other concurrent agents skip or wait instead of duplicating it.

- `scripts/agent-work-lock.js` — `claimWork(key, { agentId, ttlMs })`, `releaseWork(key, agentId, { force })`, `listClaims({ includeStale })`.
- Atomic acquisition via `fs.open(..., 'wx')` (O_EXCL) under `~/.thumbgate/runtime/work-claims/<sanitized-key>.json`. Two near-simultaneous claims resolve to exactly one winner.
- **Stale handling:** a claim is reclaimable once it is older than its TTL (default 30 min) **or** its owning PID is no longer alive — a crashed agent never permanently blocks a work unit.
- Surfaced as three **additive** MCP tools on the existing stdio server: `claim_work`, `release_work`, `list_work_claims` (following the existing `TOOLS` array + `switch`-case registration pattern; no existing handler changed).

## Why — the 2026-06-06 double-work incident

When a user runs MULTIPLE autonomous coding agents (Claude Code, Codex, Antigravity) on the same repo, two agents can pick up and work the **same** task at once → duplicate/conflicting work + wasted tokens.

This literally happened on **2026-06-06**: two agents both built the statusline feedback-aggregation fix; one (`feedback-aggregate-stats.js`) became dead code. A claim lock would have let the second agent see the work was already taken and skip it.

## Dogfooding

ThumbGate is an agent-governance product — pre-action gates for AI agents. This feature **is** a pre-action gate: claim before you act. It dogfoods ThumbGate's own pitch on ThumbGate's own multi-agent workflow.

## Tests & verification

`tests/agent-work-lock.test.js` runs against a temp `THUMBGATE_HOME` so it never touches the real `~/.thumbgate`. Covers: first claim succeeds; contended second claim by a different agent fails and returns `heldBy`; same-agent re-claim is non-double-acquiring; release frees the lock (and wrong-agent release is denied); force release; TTL-expired claim is reclaimable; dead-PID claim is reclaimable within TTL; two near-simultaneous claims → exactly one acquires; `listClaims` reports live vs reclaimable; key-collision safety.

```
$ node --test tests/agent-work-lock.test.js
ℹ tests 12
ℹ pass 12
ℹ fail 0

$ node --test tests/test-suite-parity.test.js
ℹ tests 5
ℹ pass 5
ℹ fail 0
```

Wired `test:agent-work-lock` into `package.json` and chained into the main `test` script so the parity guard stays green. Existing tool-registry, server-lock, MCP annotation, changeset-check, bundle-ratchet, and public-core-boundary suites remain green. Changeset added (`thumbgate`: minor).

Entrypoint check uses the path-based `process.argv[1]` form, not `require.main === module` (SonarCloud S3403).


IgorGanapolsky:
Closing. The intent (stop concurrent agents from double-working) is a real failure I hit this session, but the fix belongs in operations, not the product runtime: it adds 518 lines of coordination infra into the production MCP stdio server (adapters/mcp/server-stdio.js) for a problem better solved by giving each agent its own git worktree and scoping its tools. Not shipping agent-orchestration plumbing into the customer-facing server. The operational lesson is already captured; reopen if we ever need an in-product lock for a customer reason.

PR #2589, opened 2026-06-08
fix(health): report the actually-deployed commit (RAILWAY_GIT_COMMIT_SHA)

IgorGanapolsky:
## The bug (and why the deploy job goes red)
`/health` has reported `buildSha: b7b0ddac…` for weeks. I assumed that meant production was **stale** — it isn't. Production serves current code; I verified two pages added *after* that commit return HTTP 200:

```
GET /guides/agentic-web-governance  -> 200
GET /guides/database-agent-safety   -> 200   (both new since b7b0ddac)
/health buildSha                    -> b7b0ddac…   (frozen)
```

So the **code deploys fine** (Railway's native GitHub auto-deploy), but the `buildSha` label is frozen. Chain:
1. `config/build-metadata.json` is committed as a null placeholder (`{"buildSha":null}`), so the baked-file path yields nothing.
2. `.github/workflows/deploy-railway.yml` exits the variable-sync step early because `RAILWAY_SYNC_VARIABLES != "true"`, so `THUMBGATE_BUILD_SHA` is **never updated** per deploy.
3. `railway up` stamping is unreliable (it false-fails), so no fresh file gets baked either.
4. `resolveBuildMetadata` falls back to the stale `THUMBGATE_BUILD_SHA` env (last set manually ≈ `b7b0ddac`).
5. The deploy workflow's verification gate compares `/health` buildSha to the merge SHA → can never match → **every deploy run is marked failed** even though the code shipped.

## The fix
`resolveBuildMetadata` now prefers **`RAILWAY_GIT_COMMIT_SHA`** — injected by Railway per deploy on GitHub-connected services ([docs](https://docs.railway.com/reference/variables)), i.e. the commit actually serving traffic — over the drift-prone `THUMBGATE_BUILD_SHA`. A properly baked file SHA still wins when present (preserves the 2026-05-20 anti-drift precedence).

Precedence: baked file → `RAILWAY_GIT_COMMIT_SHA` → `THUMBGATE_BUILD_SHA` → null.

## Verification
- `npm run test:build-metadata` → 10/10 (added 3 precedence tests)
- `npm run test:api` → 955/955
- **Self-verifying in prod:** after merge, `/health` buildSha flips to the real commit if Railway injects the var, or is unchanged otherwise — no regression either way.

### DIFF
--- a/.changeset/health-buildsha-railway-git-commit.md
+++ b/.changeset/health-buildsha-railway-git-commit.md
@@ -1,0 +1,5 @@
+---
+"thumbgate": patch
+---
+
+fix(health): report the actually-deployed commit in `/health` `buildSha`. On the Railway GitHub-connected service the baked `config/build-metadata.json` is the committed null placeholder and `RAILWAY_SYNC_VARIABLES` is off, so `THUMBGATE_BUILD_SHA` never updated and `/health` reported a months-old commit while newer code was live — which also made the deploy workflow's build-SHA verification gate fail on every run. `resolveBuildMetadata` now reads Railway's per-deploy `RAILWAY_GIT_COMMIT_SHA` (ground truth for GitHub-connected deploys) ahead of the drift-prone `THUMBGATE_BUILD_SHA`, while a properly baked file SHA still wins when present.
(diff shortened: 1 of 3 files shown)


IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `b485f6dd5b6311efb3209ee2274cb822fcbb7f06`
**Production buildSha:** `b485f6dd5b6311efb3209ee2274cb822fcbb7f06`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","degraded":false,"version":"1.27.6","buildSha":"b485f6dd5b6311efb3209ee2274cb822fcbb7f06","uptime":30.127738877,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

PR #2602, opened 2026-06-10
chore(gate): drop dead unverified-skill loss-matrix entry; add test:wire-proof-gate script

IgorGanapolsky:
## Summary

Two evidence-backed hygiene gaps in the skill-verification gating path (found 2026-06-10). Scoped, no behavior change.

### 1. Dead `unverified-skill` loss-matrix entry — removed
`config/enforcement.json` defined `lossMatrix.falseAllow["unverified-skill"]: 160`, but **nothing emits that tag**:
- `scripts/risk-scorer.buildPatternSummary` does not produce it (it aggregates tags already on feedback rows).
- No gate rule or feedback path tags an action `unverified-skill`.
- `grep -rn "unverified-skill" scripts/ src/ config/` finds the bare tag only in a content generator (`linkedin-content-generator.js`), the cost entry itself, and one test array.

The Bayes loss matrix only has teeth for tags that appear in learned feedback/lesson data (`hook-pre-tool-use.js` → `tagsForLesson` → `resolveCost`), so the entry never participated in a real decision.

**Skill-verification gating that *does* ship is unaffected** — it runs through the separate deterministic rule `unverified-skill-use` in `config/gates/default.json` (Supply Chain layer, `Skill:load` trigger, warn-unless-`skill_verified`). That is a different subsystem and a different tag id; it is not touched here.

Removed the orphaned cost entry and its assertion in the expanded-family loss-matrix test (`tests/bayes-optimal-gate.test.js`), which was the only thing keeping the dead key "green."

### 2. `test:wire-proof-gate` script — added
`tests/wire-proof-gate.test.js` had no dedicated npm script. Added `test:wire-proof-gate` and chained it into the root `test`, matching the per-gate-test convention (`node --test tests/<name>.test.js`).

> Note for accuracy: the file already executed in CI via `test:proof:truth` (which bundles it), so it was not actually "rotting." This change gives it a **named, individually-runnable entry point** consistent with every other gate test, so it cannot silently drop out of coverage if the proof bundle is ever re-shaped.

## Verification (repo uses `node:test`, not jest)
- `test:wire-proof-gate` 1/1 · `test:bayes-optimal-gate` 44/44
- `test:enforcement-teeth` 38/38 · `test:gate-coherence` 15/15 · `test:gate-eval` 24/24
- `test:enforcement-loop` 21/21 · `test:public-bundle-ratchet` 2/2 · `test:schema` 8/8
- `sync-version --check`: 32/32 targets in sync
- pre-commit: package parity ✓, version sync ✓


IgorGanapolsky:
Stale (27d): gate hygiene chore not landed in 27d; will re-derive from current main if still needed. Closing.

PR #2612, opened 2026-06-10
fix(landing): hero lede leads with mechanism, not vibes

IgorGanapolsky:
## Why

The hero led with a vague claim — *"AI agents will always hallucinate and break things — we make their mistakes harmless"* — before the mechanism. A skeptical buyer reads "harmless" and asks **how???**, which is exactly the conversion-killer ($0/30d, the landing is the bottleneck). GSD's docs win by leading with mechanism; this brings the hero in line.

## Change

One line — the hero lede. Now leads with **what actually happens**:
> When your coding agent tries to run **rm -rf**, force-push to main, leak an API key, or DROP a production table, ThumbGate blocks that exact tool call in the PreToolUse hook — on your machine, before the shell runs it. Thumbs-down a mistake once and it becomes a prevention rule, so the agent stops repeating it.

- Mechanism-first (where/when it blocks), names real commands.
- Surfaces the actual moat (👎 → prevention rule → stops the repeat).
- Accuracy: drops the "forever" implication — promoted rules TTL-expire (90d) unless they keep firing.
- Keeps the local-first differentiator (no server/gateway/rollout).

Landing-page-claims + competitive-positioning tests stay green.

### DIFF
--- a/.changeset/hero-mechanism-first.md
+++ b/.changeset/hero-mechanism-first.md
@@ -1,0 +1,5 @@
+---
+"thumbgate": patch
+---
+
+Rewrite the landing-page hero lede to lead with the concrete mechanism instead of a vague benefit claim. Replaces "AI agents will always hallucinate… we make their mistakes harmless" with what actually happens — ThumbGate blocks the specific dangerous tool call (rm -rf, force-push to main, leaked key, DROP on prod) in the PreToolUse hook before the shell runs it, and a thumbs-down becomes a prevention rule that stops the agent repeating it.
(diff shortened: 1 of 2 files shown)


IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `f555423104ed53d8edbe3e40166c2b371cab411f`
**Production buildSha:** `f555423104ed53d8edbe3e40166c2b371cab411f`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","degraded":false,"version":"1.27.6","buildSha":"f555423104ed53d8edbe3e40166c2b371cab411f","uptime":44.971707084,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

PR #2702, opened 2026-06-23
fix(brand): restore original TG gate monogram logo to match Stripe and bust cache to v4

IgorGanapolsky:
Restores the original TG gate monogram logo (which matches the logo used in Stripe products) and cache-busts the inline SVG asset to v4.


IgorGanapolsky:
Stale (14d): logo/monogram already correct on main; cache-bust v4 no longer needed. Closing.

PR #2739, opened 2026-07-01
fix(diagnostic): ship upgraded $499 workflow-hardening buyer page

IgorGanapolsky:
## What
Ships the upgraded `/diagnostic` ($499 Workflow Hardening Diagnostic) buyer page — intake form, `#scope`/`#proof` sections, wired `/go/diagnostic` + `/checkout/pro` CTAs.

## Why
The `/diagnostic` route and an older, thinner page were already on `main` and live in prod. The richer intake-form version was stranded **untracked** in a working tree — never committed, never deployed — so buyers were hitting the weaker page. This commits the upgrade so the full offer page ships.

## Scope
Single page + changeset. No route/server change (route already on main). `public/diagnostic.html` +409/-282.

## Verify after merge + deploy
- `curl -s -o /dev/null -w "%{http_code}" https://thumbgate.ai/diagnostic` → `200`
- Body contains `#intake` form and `$499`

### DIFF
--- a/.changeset/ship-diagnostic-buyer-page.md
+++ b/.changeset/ship-diagnostic-buyer-page.md
@@ -1,0 +1,5 @@
+---
+"thumbgate": patch
+---
+
+Ship upgraded $499 Workflow Hardening Diagnostic buyer page (intake form, scope/proof sections, wired /go/diagnostic + /checkout CTAs); replaces the thinner page currently live at /diagnostic.
--- a/public/diagnostic.html
+++ b/public/diagnostic.html
@@ -1,345 +1,467 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
 <head>
-<meta charset="utf-8">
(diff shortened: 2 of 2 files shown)


IgorGanapolsky:
Closing — not needed and not safe as-is.

The deployed /diagnostic page (origin/main) is ALREADY the complete working $499 Workflow Hardening Diagnostic intake page: intake form (action=/v1/intake/workflow-sprint), planId=diagnostic, $499 CTA, passes tests/api-server.test.js:518. The 404 was a stale deploy (prod 86 commits behind main), already fixed by deploying main HEAD.

This PR shipped an untracked orphan variant (Book $499 + extra copy + data-diagnostic-payment-handoff) that conflicts with the spec test on multiple assertions. Shipping it would mean rewriting the guard to match unvetted content. A richer page should land as a vetted, spec-aligned change, not a cherry-pick.

PR #2760, opened 2026-07-05
fix: checkout dead-end — 99 visitors tried to pay, 0 could

IgorGanapolsky:
## Revenue Impact: CRITICAL

99 visitors hit /checkout/pro in the last 30 days. 0 paid.

**Root cause:** The checkout form posts to /checkout/pro?confirm=1, which calls createCheckoutSession() server-side. This function requires STRIPE_SECRET_KEY to be set as an env var on Railway. When it fails, the catch block redirects back to the interstitial — creating an infinite loop where every 'Pay' click just reloads the same page.

**Fix:** Changed the form action to point directly to the Stripe Payment Link. This bypasses the broken server-side flow entirely. Every 'Pay' click now goes straight to Stripe's hosted checkout.

**Verification:** The Stripe Payment Link (https://buy.stripe.com/8x2dR91M84r4cSd9uj3sI3f) returns HTTP 200 and is live.

**Expected impact:** 99 visitors/month were ready to pay. Even at 5% conversion that's ~5 new customers = $95/mo MRR.

### DIFF
--- a/.changeset/fix-checkout-dead-end.md
+++ b/.changeset/fix-checkout-dead-end.md
@@ -1,0 +1,5 @@
+---
+"thumbgate": patch
+---
+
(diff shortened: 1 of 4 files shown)


IgorGanapolsky:

PR #2791, opened 2026-07-09
fix(publish): ship 5 dynamically-required scripts + closure guard

IgorGanapolsky:
Root cause: published thumbgate dropped 5 scripts hard-required by shipped-but-dynamically-loaded command scripts, crashing the UserPromptSubmit hook with `Cannot find module`. Fix adds them to package.json:files (1.27.21) + a closure guard in pack-runtime-integrity.test.js (every packed file's require target must be packed). Negative-tested. 🤖 Claude Code


IgorGanapolsky:

PR #2797, opened 2026-07-09
feat(license): honor signed, expiring entitlement tokens in the Pro gate

IgorGanapolsky:
## Why (money)
CEO wants "free month → \$19/mo" trials for warm leads. Proven with evidence that the motion **could not work**:
- \`isProLicensed()\` (scripts/license.js) is what lifts rate-limit caps (rate-limiter.js:93). It only accepted **non-expiring** \`tg_pro_\`/\`tg_\` prefix keys.
- The signed \`entitlement.js\` tokens **do** expire but were never wired into \`isProLicensed()\`.
- Isolated test: a signed 30-day token → \`isProLicensed: false\` → a "trial" unlocked nothing.

## What
\`verifyLicense()\` gains a third source after env/file prefix keys: **signed entitlement tokens** (via \`THUMBGATE_LICENSE\` or the license file), verified through \`entitlement.verifyLicense()\`, granting Pro only for \`pro\`/\`team\`/\`enterprise\` tiers. Expiry is automatic — past \`exp\`, verification fails and the caller reverts to free. **Backward compatible**: existing prefix keys resolve first, unchanged.

## Tests (evidence)
- +5 tests in \`tests/license.test.js\`: signed unexpired grants Pro; **expired reverts to free**; free-tier signed token denied; token from license file; tampered token denied. Uses an ephemeral ed25519 keypair (no dependency on the production signing key).
- Scrubbed-env (CI-like) run: **license 12/12, rate-limiter 11/11**. (Local rate-limiter failures were pre-existing on main — this session's \`THUMBGATE_API_KEY\` contaminates \`isProLicensed()\`; they pass with env scrubbed.)

### DIFF
--- a/.changeset/expiring-pro-trials.md
+++ b/.changeset/expiring-pro-trials.md
@@ -1,0 +1,5 @@
+---
+"thumbgate": minor
+---
+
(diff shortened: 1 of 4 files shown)


IgorGanapolsky:

PR #2799, opened 2026-07-09
feat(rules): surface leverage points by clustering related prevention rules

IgorGanapolsky:
## Problem
Dedup (lesson-synthesis) merges *identical* mistakes, so N distinct symptoms of one habit become N narrow gates — rule sprawl, and the single high-leverage fix is never surfaced.

## What
`scripts/rule-clustering.js` groups related-but-distinct rules into **leverage points** in the feedback rules report:
- Related if they share a significant tag OR token-overlap ≥ 0.30 (a band **below** the 0.6 dedup threshold — catches distinct-but-related, not identical).
- Union-find → clusters; each carries the **shared tags/terms as evidence**.
- `surfaceLeveragePoints()` returns clusters ≥ 3 members with a human-confirmable suggestion; rendered in `feedback:rules` output.

## Deliberately NOT a causal graph
Makes **no** causal claim — groups by lexical/tag relatedness, shows the evidence, and every suggestion says "candidate cluster, confirm before acting." Causal attribution is a separate interpretive layer, out of scope by design.

## Tests
- `tests/rule-clustering.test.js` (8/8): 5 distinct symptoms of one habit → 1 leverage point; unrelated rules stay separate; token-only relatedness; determinism; missing-tags/empty input; auditable markdown.
- `feedback-to-rules` integration: 6/6.


IgorGanapolsky:

PR #2807, opened 2026-07-10
fix(gates): stop deny-network-egress from matching every curl/wget

IgorGanapolsky:
## Problem

`deny-network-egress` led with bare `curl\s|wget\s` alternatives:

```
curl\s|wget\s|fetch\(|https?://(?!github\.com|registry\.npmjs\.org|api\.anthropic\.com)
```

Because `curl\s` is its own alternative, **any** `curl` or `wget` matched regardless of destination — the allowlist was never consulted for them. `curl http://localhost:9222` (a loopback CDP probe) warned on every single invocation, and `curl https://github.com/...` warned despite `github.com` being allowlisted.

Those false positives are recorded as negative feedback events. On one install the gate had warned 457 times, and `gate:deny-network-egress` sat at 48/48 negative — feeding a 1.0 risk-rate back into the session-start "top high-risk tags" banner. The gate was training the model on its own exhaust.

## Second, quieter bug

Each allowlist entry matched as a bare hostname **prefix** with no boundary, so:

```
open https://github.com.evil.com/x          -> exempt
open https://api.anthropic.com.attacker.net -> exempt
const u = 'https://github.com.evil.com/p'   -> exempt
```

This was only reachable for non-`curl` invocations, because the blanket `curl\s` alternative masked it whenever a curl token happened to be present. Removing `curl\s` would have exposed it, so both are fixed together.

## Fix

Detection is now driven by the **destination**, not the binary name:

| alternative | behavior |
|---|---|
| `fetch(` | unchanged |
| non-allowlisted http(s) URL | allowlist extended with loopback (`localhost`, `127.0.0.1`, `[::1]`, `0.0.0.0`) |
| `curl`/`wget` at a bare host | preserves coverage the URL rule would miss (`curl example.com`) |

Every allowlist entry is anchored to a hostname boundary `(?:[:/?#]|$)`.

`action`, `layer`, `severity`, `unless`, and `compliance` are unchanged — this is warn-level, same as before.

## Tests

Adds `tests/deny-network-egress-pattern.test.js` (wired into `test:gates`), covering loopback, allowlisted hosts, real egress, and the prefix-bypass regression. **The new test fails against the previous pattern** (verified by reverting the pattern in place and re-running).

- `node --test tests/deny-network-egress-pattern.test.js` — 5/5
- `npm run test:gates` — 283/283
- pre-push regression guard (`public-package-parity`, `test-suite-parity`, `landing-page-claims`) — 33/33
- `test:gate-coherence`, `test:gate-eval`, `test:gates-hardening` — pass

## Note on existing data

This stops *new* false negatives. It does not retroactively clean feedback rows already recorded from the false positives; those still skew tag risk-rates and the approval rate until purged or down-weighted.


IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `811635b94e885f6b160ff1cb7508455fb1d17920`
**Production buildSha:** `811635b94e885f6b160ff1cb7508455fb1d17920`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","degraded":false,"version":"1.27.20","buildSha":"811635b94e885f6b160ff1cb7508455fb1d17920","uptime":27.530796163,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

PR #2809, opened 2026-07-10
fix(publish): bound the no-op guard's pending-changeset exemption

IgorGanapolsky:
## The bug

`publish-npm.yml` already errors when shipped content changed since the last release tag but `package.json` was never bumped. It exempts that error whenever **any** pending changeset exists, on the theory that "the next versioned publish will ship it."

Nothing bounded *until*.

On 2026-07-09, PR #2807 (the `deny-network-egress` host-boundary bypass) merged as `811635b`. `Publish to NPM` ran for that commit and reported **success while publishing nothing**, because a changeset existed:

```
::notice::Shipped content changed since v1.27.20, but 9 pending changeset file(s)
were detected. Treating this no-op as release-audited until the next versioned publish lands.

Report publish no-op:
  Version 1.27.20 is already published from commit adfbcec8. Skip npm publish for this merge…
```

I confirmed the consequence by pulling `thumbgate-1.27.20.tgz` from the registry and unpacking it — **the published package still contains the vulnerable pattern**:

```
published:   curl\s|wget\s|fetch\(|https?://(?!github\.com|registry\.npmjs\.org|api\.anthropic\.com)
origin/main: fetch\(|https?://(?!(?:github\.com|…|0\.0\.0\.0)(?:[:/?#]|$))|(?:curl|wget)\s+…
```

A merged security fix, a green pipeline, and an unpatched package — all true at once. **An exemption with no expiry is a silence, not an audit.**

## The change

The exemption now lapses after `MAX_UNRELEASED_DAYS` (default 7). Past that, a pending changeset stops excusing the no-op and the run fails, pointing at the release PR.

The decision moves out of inline shell into `scripts/release-window.js` so it can be tested. `tests/release-window.test.js` pins the boundary and the 2026-07-09 regression.

## Verification

A/B against the rule it replaces, on the real incident's numbers (`shipped=12`, `pending=9`, last release `2026-07-08T20:47:35Z`):

| scenario | OLD rule | NEW rule |
|---|---|---|
| merge 4h after release | PASS | PASS |
| merge 2 days after | PASS | PASS |
| merge 14 days after | PASS | **FAIL** ← newly caught |
| merge 60 days after | PASS | **FAIL** ← newly caught |

The old rule passes every scenario, including a security fix unshipped for two months. Normal development is unaffected.

**Checked against `main` as it stands today:** last tag `v1.27.20` is 1 day old, so the guard exits 0 and `main` stays green. It fires on day 8 unless a release is cut. This PR does not turn `main` red.

`8/8` unit tests pass. `bash -n` clean on the guard step; workflow YAML parses. Registered in `npm test` — the repo's own `test-suite-parity` guard caught that omission on first push, which is exactly what it's for.

## Not in scope

This does not ship #2807. That still needs a `release:` PR (bump + consume changesets). Do not hand-publish — 1.27.19 shipped broken that way.


IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `e216172aacf9724af0e389fb2f73f4408b6535bb`
**Production buildSha:** `e216172aacf9724af0e389fb2f73f4408b6535bb`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","degraded":false,"version":"1.27.20","buildSha":"e216172aacf9724af0e389fb2f73f4408b6535bb","uptime":40.137852004,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

PR #2849, opened 2026-07-10
feat(gates): self-protection gates bind regardless of enforcement posture

IgorGanapolsky:
## The gap this closes

ThumbGate ships **warn-by-default** so legitimate work is never hard-blocked. But that downgraded *every* deny to a warning — including the gates guarding ThumbGate's own guardrails. An agent could:

```
pkill -f gates-engine        # kill the enforcement engine
export THUMBGATE_HOTFIX_BYPASS=1   # disable all checks
```

…and receive only a warning. By the time it's logged, the guardrail is gone. That's the one class of action where a warning is worthless — the exact criticism that "it watches, it doesn't stop."

## The change

`applyEnforcementPosture` now keeps the `deny` for self-protect gates (`gateId.startsWith('self-protect')`) regardless of posture. Verified via the real `gate-check` subprocess:

| action | before | after |
|---|---|---|
| `pkill -f gates-engine` | warn | **deny** |
| `export THUMBGATE_*=…` | warn | **deny** |
| `git push --force main` | warn | warn (unchanged) |

The last row is the point: this is **targeted, not a blanket hard-block**. The CEO's warn-by-default policy still applies to everything that isn't self-protective.

## Escape hatch (no self-lockout)

`THUMBGATE_SELF_PROTECT_OVERRIDE=1` downgrades self-protect back to warn, so an accidental warn posture can never strand the owner from repairing their own gate config (self-lockout, 2026-07-07). It's a *distinct* escape because break-glass covers `.claude/settings*` but not the self-protect surface.

## Tests

`tests/self-protect-enforcement.test.js` runs the real subprocess. Control-verified: reverting the `gates-engine` change fails the two deny assertions; restoring → 7/7. **189/189** existing gate-engine tests still pass.

## Scope (honest boundary)

The two self-protect gates that fire *first* — `self-protect-kill`, `self-protect-env-override` — now hard-deny. The file-edit gates (`self-protect-config`, `self-protect-hooks-disable`) are shadowed by `protected-file-approval-required` / `workflow-sentinel` and remain approval-gated. Hardening those to hard-deny requires gate reordering in a hot path — a tracked follow-up, deliberately **not** bundled into this release.


IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `f49cd34879526b9926c02ba9301f4ae0983f44be`
**Production buildSha:** `f49cd34879526b9926c02ba9301f4ae0983f44be`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","degraded":false,"version":"1.27.20","buildSha":"f49cd34879526b9926c02ba9301f4ae0983f44be","uptime":17.159039032,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

PR #2862, opened 2026-07-11
feat(gates): catastrophic-floor deny for truly irreversible gates (force-push, rm -rf home/root)

IgorGanapolsky:
## What

`applyEnforcementPosture()` (warn-by-default, CEO 2026-06-04) downgraded every `deny`/`approve` to `warn` except self-protect gates. Above it sat **two contradictory comment blocks**:

1. One claimed rm-rf-class stays a hard deny "via `DESTRUCTIVE_FS_PATTERN` below" — **that constant is never defined anywhere in the file** (false comment).
2. The next explained the team had *deliberately rejected* a regex catastrophic floor as evadable theater.

So the documented "catastrophic floor" did not exist: `rm -rf /` and `git push --force` both silently downgraded to `warn` by default.

This PR adds a real, minimal catastrophic floor implemented as a **gate-id exemption** — reusing the engine's existing gate detection (the same mechanism `self-protect` already uses), **not** a new regex over the raw command (which `sudo` / `bash -c` / `find -exec` / `base64|sh` all evade).

## Changes

- `CATASTROPHIC_GATE_IDS = { 'force-push', 'rm-rf-home-or-root' }` + `isCatastrophicGate()` helper. Deliberately tiny: only genuinely irreversible actions (force-push rewrites published history; rm -rf of `/` `~` `$HOME` wipes the filesystem). Everything recoverable (git revert, reflog, redeploy, `rm -rf node_modules`) still warns.
- In `applyEnforcementPosture`, after the self-protect check, keep `deny` for catastrophic gate ids. Owner escape `THUMBGATE_CATASTROPHIC_OVERRIDE=1` mirrors `THUMBGATE_SELF_PROTECT_OVERRIDE=1`.
- Reconciled the two contradictory comment blocks into one accurate comment and deleted the phantom `DESTRUCTIVE_FS_PATTERN` reference.
- Exported `applyEnforcementPosture` / `isCatastrophicGate` / `CATASTROPHIC_GATE_IDS` for unit testing.

## Why this shape

- Force-push is intercepted by the `local-only-remote-side-effect` scope gate when a task scope is active, so a subprocess assertion for it is environment-dependent. It is therefore proven at the unit level against `applyEnforcementPosture` directly; `rm -rf` is proven end-to-end through the real `bin/cli.js gate-check` subprocess (never shadowed).

## Test evidence

New `tests/enforcement-catastrophic-floor.test.js` (wired into `npm test` + passes the `test-suite-parity` guard). Demonstrated FAIL-before / PASS-after:

- Against unmodified engine: `# pass 2 # fail 6` (only the two CONTROL cases pass — floor absent).
- After the change: `# pass 8 # fail 0`.

Controls prove the floor is **targeted, not a blanket block**: a non-catastrophic deny (`git reset --hard`) still warns by default; `THUMBGATE_CATASTROPHIC_OVERRIDE=1` downgrades the catastrophic deny to warn.

Reconciled existing suites (both green):
- `tests/gates-engine.test.js` — updated the `rm -rf /` case to assert deny-by-default + override + strict (175 pass / 0 fail).
- `tests/self-protect-enforcement.test.js` — its CONTROL used force-push (now catastrophic); switched to `git reset --hard` (still recoverable → warn) (7 pass / 0 fail).


IgorGanapolsky:

PR #2928, opened 2026-07-14
fix(mcp): never reap a live MCP server on lock age alone — use heartbeat staleness

IgorGanapolsky:
## Problem

Any ThumbGate MCP server session older than 2 hours was SIGTERM'd by the next agent session that started against the same feedback dir. The stale-lock reaper in `adapters/mcp/server-stdio.js` treated lock **age** (`startedAt`) as proof of orphaning, but the lock was written once and never refreshed — so a live, actively-serving server looked identical to a wedged orphan.

Observed 2026-07-13: a long-running agent session's `thumbgate.recall` failed with `Transport closed` because a newly started session reaped its server mid-call. Matches the previously captured lesson: "age alone was treated as proof of orphaning, which can close an active client's stdio transport" (live PID 90160 reaped).

## Fix

- Lock holders now refresh a `heartbeatAt` field every `LOCK_HEARTBEAT_MS` (default 60s, `THUMBGATE_LOCK_HEARTBEAT_MS` override), via an unref'd timer.
- The reaper fires only when the **heartbeat** is older than `LOCK_STALE_MS`. Healthy servers are never reaped regardless of session length.
- Wedged processes (event loop dead → no heartbeat) and pre-heartbeat orphans (no `heartbeatAt`, stale `startedAt`) are still reclaimed — the original orphan protection is preserved.
- `cleanupLock` stops the heartbeat timer before unlinking, so a straggling tick can't resurrect a removed lock.

## Evidence

End-to-end before/after with the real server binary (server A up 3.5s, threshold 2s, server B starts):

```
=== OLD (origin/main) ===
[B STDERR] [thumbgate] Lock held by PID 94919 is 0m old (threshold: 0m). Reaping orphaned process.
[A EXIT] code=0 signal=null
A alive after B started: false
A still serving requests: false

=== PATCHED ===
[B STDERR] [thumbgate] Another MCP server (PID 96212) is running ... Starting concurrent session.
A alive after B started: true
A still serving requests: true
```

Tests: `tests/server-stdio-lock.test.js` 9/9 pass (7 existing + 2 new: fresh-heartbeat holder is not killed; heartbeat advances and cleanup stops it). Related suites `mcp-server`, `recall-limit`, `mcp-tools-gates`: 64/64 pass.

https://claude.ai/code/session_01Ur1G48mqfyU8u8zj52suvU


IgorGanapolsky:
Closing as a duplicate of #2929. This heartbeat variant still reaps live pre-heartbeat servers after the age threshold, documents a host-sleep false-reap residual, and carries unrelated AGENTS.md/GEMINI.md changes. #2929 applies the narrower invariant: never terminate a live stdio owner solely from lock age.

PR #2962, opened 2026-07-20
feat(web): add Leash cloud-control product page

IgorGanapolsky:
## Summary
- add the public `/leash` product page for free local web control and paid managed failover
- add an environment-controlled, privacy-safe bridge to the authenticated control plane
- publish SEO/GEO discovery through the sitemap, JSON-LD, and LLM manifests

## Verification
- `node --test tests/public-package-parity.test.js tests/test-suite-parity.test.js tests/landing-page-claims.test.js` — 33/33 pass
- `npm run test:public-package-parity` — 9/9 pass
- `npm run test:public-static-assets` — 87/87 pass
- `npm run verify:quick` — 281/281 pass
- repository pre-push guard — pass
- real Playwright render inspected locally

PR #2965, opened 2026-07-21
fix(deps): patch protobufjs/js-yaml/adm-zip transitive vulnerabilities

IgorGanapolsky:
## Summary
- \`npm audit\` reported 21 vulnerabilities, currently failing the "Audit root npm dependencies" step in the \`test\` CI job on **every open PR** (confirmed identically failing on unrelated PR #2962).
- Root cause: existing \`overrides\` for \`protobufjs\` (pinned \`7.6.4\`) and \`js-yaml\` (pinned \`4.2.0\`) were both sitting *inside* their own vulnerable ranges (\`7.5.0-7.6.4\` and \`4.0.0-4.2.0\`), and \`adm-zip\` resolved to the vulnerable \`0.5.17\` via \`onnxruntime-node\`.
- Fix: minimum safe patch bump within the same major version for all three — \`protobufjs\` 7.6.4→7.6.5 (overrides) / 8.5.0→8.7.0 (top-level dep), \`js-yaml\` 4.2.0→4.3.0, \`adm-zip\` 0.5.17→0.6.0. No major-version jumps, no breaking API changes, so \`@huggingface/transformers\`/\`onnxruntime-web\`/\`@google/genai\` themselves are untouched.
- Down to 1 remaining vuln (\`brace-expansion\`) — already covered by open dependabot PR #2963, intentionally not duplicated here to avoid a merge conflict between the two.

## Test plan
- [x] \`npm audit\` — 21 → 1 vulnerability (the dependabot-owned one)
- [x] Verified \`js-yaml\` load/dump, \`adm-zip\`, \`protobufjs.Root\` all still load correctly at the bumped versions
- [x] \`tests/oss-pr-opportunity-scout.test.js\` + \`tests/upstream-contribution-engine.test.js\` (the two suites exercising this dependency chain) — 17/17 pass
- [x] \`node scripts/changeset-check.js\` — no changeset required (dependency-only change)
- [x] \`node scripts/sync-version.js --check\` — 38/38 targets in sync

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
https://claude.ai/code/session_013CYitxTcqYAab6352rdXgP

### DIFF
--- a/.changeset/patch-transitive-audit-vulnerabilities.md
+++ b/.changeset/patch-transitive-audit-vulnerabilities.md
@@ -1,0 +1,5 @@
+---
+"thumbgate": patch
+---
+
+Patch protobufjs, js-yaml, adm-zip, and brace-expansion transitive dependency versions via npm overrides to clear all npm audit vulnerabilities, with no major-version bumps to any directly-used package.
--- a/.changeset/patch-transitive-audit-vulnerabilities.md
+++ b/.changeset/patch-transitive-audit-vulnerabilities.md
@@ -2,4 +2,4 @@
 "thumbgate": patch
 ---
 
-Patch protobufjs, js-yaml, adm-zip, and brace-expansion transitive dependency versions via npm overrides to clear all npm audit vulnerabilities, with no major-version bumps to any directly-used package.
+Patch protobufjs, js-yaml, adm-zip, and brace-expansion transitive dependency versions to clear npm audit for this repo's own root install (dev/CI), with no major-version bumps to any directly-used package. protobufjs is fixed for consumers too via a direct-dependency bump; js-yaml/brace-expansion only ever affected dev tooling. adm-zip (pulled in transitively via onnxruntime-node, install-time only, DoS-class) remains unfixed for consumers of the published package since npm overrides do not propagate past the root project — tracked as a follow-up pending an upstream onnxruntime-node/@huggingface/transformers bump.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1554,9 +1554,9 @@
       "license": "MIT"
     },
     "node_modules/brace-expansion": {
-      "version": "5.0.6",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.6.tgz",
-      "integrity": "sha512-kLpxurY4Z4r9sgMsyG0Z9uzsBlgiU/EFKhj/h91/8yHu0edo7XuixOIH3VcJ8kkxs6/jPzoI6U9Vj3WqbMQ94g==",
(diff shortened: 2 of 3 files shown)


IgorGanapolsky:
Verified this empirically with a real \`npm pack\` tarball installed as a nested dependency of a clean consumer project (not just a synthetic repro) — you're right, and more precisely than I first assumed:

- **protobufjs**: actually fine for consumers — it's a real top-level \`dependencies\` entry (not just an override), so it installs and dedupes correctly regardless of root/nested position. Confirmed resolved at 8.7.1 in the consumer install.
- **js-yaml, brace-expansion**: only ever \`devDependencies\` (changesets/lint tooling) — never shipped to consumers at all, override or not. Confirmed absent from the consumer install's \`node_modules\`.
- **adm-zip**: confirmed **not** fixed for consumers. \`overrides\` only apply when the package is the install root — this repo's root/CI is fixed, but \`npm install thumbgate\` in someone else's project still resolves the vulnerable 0.5.17 via \`onnxruntime-node\`'s own \`^0.5.16\` pin, which structurally excludes the patched 0.6.0 (0.x semver treats a minor bump as breaking).

Severity context: the underlying CVE (GHSA-xcpc-8h2w-3j85, CVSS 7.5) is a DoS via a crafted zip triggering 4GB allocation. \`onnxruntime-node\` only invokes \`adm-zip\` in its own postinstall to extract a prebuilt binary from the npm registry — not on any thumbgate runtime path that processes user/agent-controlled zip data. Real gap, low practical exploitability for this product's actual usage.

Closing this gap fully requires either an upstream \`onnxruntime-node\`/\`@huggingface/transformers\` dependency bump (not in our control) or a postinstall-patch workaround (real added complexity/risk, out of scope for what this PR set out to fix — the root/CI audit gate blocking every open PR). Updated the changeset to state this precisely rather than claim a full fix. Tracking the consumer-facing adm-zip gap as a follow-up.

Resolving this thread — the finding was correct and is now accurately reflected, not swept under the rug.

IgorGanapolsky:
Superseded — PR #2967 (already merged to main) independently fixed this exact same npm-audit blocker (same 4 packages: protobufjs, js-yaml, adm-zip, brace-expansion), which is why this PR now shows a real merge conflict rather than the earlier opaque Trunk-queue failure. Confirmed \`main\` is now at 0 vulnerabilities.

Note: the consumer-facing adm-zip gap I found and filed as #2968 still applies to #2967's fix too (same override-based approach, same npm limitation — overrides don't propagate to nested/consumer installs). That issue stands regardless of which PR landed the root/CI fix.

Closing in favor of #2967 rather than resolving a redundant conflict.

PR #2978, opened 2026-07-22
fix(ci): allowlist gate for npm audit findings with no available fix

IgorGanapolsky:
## Summary
Discovered while debugging an unrelated Stop-hook PR: `npm audit --audit-level=low` currently fails CI for every PR against main, including a fresh, untouched checkout of main itself. Root cause is `GHSA-f88m-g3jw-g9cj` (sharp <0.35.0 inherited libvips CVEs, via `@huggingface/transformers`). `npm audit` reports `fixAvailable: false` for both -- there is no upgrade path yet.

- Replaces the raw `npm audit --audit-level=low` step with `scripts/npm-audit-gate.js`, which still fails the build on **any** advisory URL not explicitly allowlisted.
- `.audit-allowlist.json` allowlists exactly this one advisory, with a written reason and a `reviewBy: 2026-08-22` date to force reassessment once a fix ships.
- `sharp` itself is directly used (not just transitively pulled in) in `scripts/social-analytics/generate-instagram-card.js` and `generate-applyops-instagram-card.js`, both of which render self-generated SVG templates, not external/untrusted image input -- noted in the allowlist reason for anyone reassessing risk later.

## Test plan
- [x] `node --test tests/npm-audit-gate.test.js` -- 7/7 pass, including a fixture test that proves the gate still rejects an unlisted advisory (not a rubber stamp)
- [x] Ran the gate against the real, live npm audit output (real `npm ci`, not a synthetic fixture) -- passes with the real allowlist, fails with an empty one
- [x] `node --test tests/test-suite-parity.test.js` -- 5/5 pass (new test file wired into the `test` chain)
- [x] `node scripts/sync-version.js --check` -- in sync

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01SQF2NhwjQy7FcVgi4Ki27T


IgorGanapolsky:
Closing as unnecessary: #2976 (merged 2026-07-22T02:44:38Z) removed the vulnerable sharp/@huggingface/transformers dependency path entirely, so `npm audit --audit-level=low` is now genuinely clean on main (verified: 0 vulnerabilities via a fresh `npm ci` against 2f6a8153). The allowlist-gate mechanism built here is no longer needed to fix that blocker — merging it now would just be unused infrastructure for a problem that no longer exists. Keeping this branch around isn't worth it either; the diagnosis work (identifying the CVE, confirming it blocked every PR) is preserved in this PR's history if the sharp/libvips advisory ever resurfaces.

PR #3013, opened 2026-07-23
fix(web): Pro $19/mo + Enterprise $499 cash paths on homepage

IgorGanapolsky:
## Summary
- Restore **both** paid paths buyers need: self-serve **Pro at $19/mo** and the **$499 Enterprise Workflow Gate**.
- Homepage nav: primary `Pro · $19/mo` → `/checkout/pro`; secondary `Enterprise · $499` → managed form.
- Hero: dual offer cards (Pro self-serve + $499 checkout form).
- `/pricing`: same dual-offer layout.
- Congruence + landing tests updated so CI requires both paths (no more single-$499 gate).

## Why
Production was $499-only. That hid the self-serve product and forced every visitor into enterprise pricing.

## Test plan
- [x] `node --test tests/public-landing.test.js tests/landing-page-claims.test.js`
- [x] `node scripts/check-congruence.js`
- [ ] After merge: curl production for `$19` + `$499` + `/checkout/pro`

### DIFF
--- a/.changeset/homepage-pro-and-enterprise-ctas.md
+++ b/.changeset/homepage-pro-and-enterprise-ctas.md
@@ -1,0 +1,5 @@
+---
+"thumbgate": patch
+---
+
+Restore both cash paths on homepage and pricing: self-serve Pro at $19/mo plus the $499 enterprise workflow gate, so operators are not forced into a single enterprise CTA.
--- a/config/github-about.json
+++ b/config/github-about.json
@@ -3,7 +3,6 @@
   "repositoryUrl": "https://github.com/IgorGanapolsky/ThumbGate",
   "homepageUrl": "https://thumbgate.ai",
   "githubDescription": "ThumbGate Pre-Action Checks self-improve from ranked lessons and repeated failures, hard-block detected secret leaks, and block matches in strict mode.",
-  "metaDescription": "ThumbGate is the self-improving firewall for AI agents under your control. Pro is $19/mo self-serve; the $499 enterprise entry installs one hard, test-backed safety gate for a supported AI-agent workflow in two business days.",
   "topics": [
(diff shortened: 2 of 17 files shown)


IgorGanapolsky:
FYI — I built and closed a duplicate PR (#3014) for this same CEO directive before noticing this one was already open and further along. In the process I found the exact cause of all 3 current \`test\` failures, in case it saves a discovery cycle:

1. **\`tests/public-static-assets.test.js:112\`** (\"landing page does not render empty revenue links\") and **\`:127\`** (\"landing page presents one fixed-price managed offer clearly\") — both assert \`doesNotMatch(html, /\/checkout\/pro|\/go\/sprint|.../)\` and \`doesNotMatch(html, /\$19|\$149|.../)\` against the live-rendered \`/\` route. These are separate from \`landing-page-claims.test.js\` and \`public-landing.test.js\` (already in your diff) — a 9th/10th independent enforcement layer I only found by running the *full* \`npm test\`, not just the targeted files. Same fix pattern as your other test updates: allow \`/checkout/pro\` as a non-primary link, allow \`$19\`.

2. **\`tests/public-enterprise-capability-boundary.test.js:66\`** (\"proposal-only expansion prices remain internal catalog truth\") — expects the literal string \`/one offer · one workflow · one fixed price/i\` in \`pricing.html\`'s eyebrow/title. Your new title (\"Pricing — Pro \$19/mo and Enterprise Gate \$499\") no longer contains that phrase, which makes sense given the dual-offer redesign — this assertion needs updating to match the new single-vs-dual-offer framing, not reverted.

Not pushing to your branch since it's actively in flight — just flagging so whoever picks this up next doesn't have to re-discover it. My closed PR (#3014) has the exact working fix for failure #1's pattern if useful as reference.

PR #3014, opened 2026-07-23
feat(pricing): restore $19/mo Pro as a secondary self-serve offer

IgorGanapolsky:
## Summary
Restores the $19/mo Pro self-serve option, unlinked since PR #2999, as a secondary CTA on `public/index.html` and `public/pricing.html` — visible and reachable via `/checkout/pro` (already a working, live checkout flow — never actually deleted), but strictly secondary to the $499 Enterprise Workflow Gate, which stays the first/primary offer on both pages.

## Why
CEO direction 2026-07-23: "we need the $499 enterprise option, but the $19 pro option as well." This directly reverses PR #2999 ("make the $499 enterprise entry unmistakable"), which explicitly removed Pro from the primary conversion path, citing near-zero self-serve conversion (documented elsewhere as 2,252 checkout sessions, 1 external completion — 0.04%). That data context is real and worth keeping in mind, but the CEO gave a clear, direct, repeated instruction to restore Pro's visibility, so this PR does that — carefully, as a secondary offer, not a full revert to the old multi-tier layout.

## What changed
- `public/index.html`, `public/pricing.html`: new `.pro-alt-offer` link inside/near the $499 checkout card — "Not ready for a $499 white-glove workflow review? Start Pro self-serve — $19/mo" → `/checkout/pro`.
- `scripts/check-congruence.js`: the single-cash-path enforcement now checks *ordering* ($499 must appear before `/checkout/pro`, Pro must never be a primary form `action`) instead of rejecting `/checkout/pro` outright.
- `docs/MARKETING_COPY_CONGRUENCE.md`: documents the reversal and the new placement policy.
- 5 test files updated to match (discovered via 3 separate enforcement layers — `node --test`, the pre-commit `landing-page-claims` hook, and `check-congruence.js` itself — each had independently pinned "Pro must be absent"):
  - `tests/positioning-contract.test.js`
  - `tests/public-landing.test.js`
  - `tests/pricing-page-telemetry.test.js`
  - `tests/version-metadata.test.js`
  - `tests/landing-page-claims.test.js`

## Test plan
- [x] All 5 updated test files pass individually and together (122+ assertions across positioning/landing/pricing/claims suites)
- [x] `node scripts/check-congruence.js` passes clean
- [x] Pre-commit hook's `landing-page-claims` guard passes clean (this is what caught the 4th, previously-undiscovered enforcement layer)
- [x] Pre-push guards (npm pack dry-run, internal link validation, regression-guard) all pass
- [x] Verified `/checkout/pro` is a real, live, working Stripe checkout flow (curl'd production: title "Confirm - ThumbGate Pro", mentions $19 and Stripe) — not rebuilding anything, just re-linking

https://claude.ai/code/session_01A2W8sjquSF36xGRcjuBfGF

### DIFF
--- a/.changeset/restore-pro-19-secondary-offer.md
+++ b/.changeset/restore-pro-19-secondary-offer.md
@@ -1,0 +1,5 @@
+---
+thumbgate: patch
+---
+
+Restore the $19/mo Pro self-serve option as a secondary link on the homepage and pricing page, alongside the $499 Enterprise Workflow Gate. PR #2999 removed Pro from the primary conversion path entirely (citing near-zero self-serve conversion); the CEO reversed that decision 2026-07-23 — Pro must be visible and reachable again, but strictly secondary to the $499 primary offer, which stays first. Updates `scripts/check-congruence.js`'s single-cash-path enforcement to allow this ordering instead of rejecting `/checkout/pro` outright.
--- a/docs/MARKETING_COPY_CONGRUENCE.md
+++ b/docs/MARKETING_COPY_CONGRUENCE.md
@@ -29,6 +29,8 @@
 **Title:** ThumbGate Pro — Personal Local Dashboard + DPO Export
 **Description:** ThumbGate Pro gives individual operators a personal local dashboard, DPO export, advanced data exports, and review-ready workflow support. Enterprise remains intake-first for scoped shared-governance requirements; hosted team lesson sync and a hosted org dashboard are not general-availability runtime features.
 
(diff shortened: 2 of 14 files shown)


IgorGanapolsky:
Closing as a duplicate of #3013, opened earlier (13:53:30Z vs my later push) by a parallel session solving the same CEO directive (restore $19/mo Pro alongside $499 Enterprise). #3013's approach is more complete — a full co-equal `.pro-card` + dedicated nav CTA, not just a secondary link — and is already further along in CI. Deferring to it rather than racing two competing implementations of the same change to the same files.

PR #3093, opened 2026-07-29
fix(rag): audit remediation wave 1 — attribution at capture, honest embeddings, safe intake

IgorGanapolsky:
Eight fixes grounded in the 6-agent RAG audit, each tested, with negative controls where the fix could be faked. Headline: feedback now derives `lastAction` from the audit trail at capture time — the attribution field existed since #203 and was populated **0 times in 1,793 entries**, which is why no retrieval golden set could ever be built. Labels now accrue from live use. Also: reranker substring-expansion poisoning fixed ('format' no longer injects delete/destroy), embedder availability made honest per its own contract, transformers declared as optionalDependency, embedding cache bound to dimensionality, symlink escape closed on both import guards, and the 201st-import catalog eviction fixed (first attempt was silently re-clamped — persistence no longer shares the search code path).

https://claude.ai/code/session_01MXUnhsxB1QNsYFKrmAunEB


IgorGanapolsky:

PR #3104, opened 2026-07-30
fix(surface): make the declared surface a generated artifact of the shipped one

IgorGanapolsky:
Two external failures last night — Anthropic's directory review (listing declares 82 tools, server exposes 42) and Glama's container build — share one root cause with most of this week's internal bugs: **a declared surface drifting from the actual surface, with nothing comparing them.**

- **Tool list is now generated** from the live server (`npm run manifest:export` → `config/directory-manifest.json`) with a CI parity test. Adding/removing a tool now fails the build until the listing is regenerated. It also derives conversation-data tools from their schemas and asserts the disclosure matches — finding `require_evidence_for_claim`, which the directory review itself missed.
- **We now build our own Dockerfile** (arm64 + amd64) and prove the image answers `/health`. Nothing built it before; third-party builders were our only test.
- **Native modules → optionalDependencies.** Both already load via `loadOptionalModule`; verified the server starts and `/health` returns 200 with both deleted.

Not claimed: that this fixes Glama's build (their log is behind auth). It removes the most probable cause and makes that failure reproducible in our CI.

https://claude.ai/code/session_01MXUnhsxB1QNsYFKrmAunEB


IgorGanapolsky:

PR #3119, opened 2026-07-31
fix(gates): make auto-promoted gates actually enforce

IgorGanapolsky:
## Problem

The thumbs-down → block loop was inert. 3× 👎 on a command produced a gate that rendered as `action:"block"`, `severity:"critical"`, `occurrences:3` — while `gate-check` on that exact command still returned **allow**.

This fails in the worst direction: the dashboard shows a critical blocking rule, so it reads as "the agent learned," while the next identical command runs.

This is the mechanism behind the **"Self-Improving Firewall"** positioning. Verified inert on `v1.30.0` / `origin/main`.

## Two defects

**1. The gate's match pattern was the group key.** `buildGateRule` used `group.key` as `pattern`. Keys are usually tag-derived (a `kubectl` capture got `["entity:Customer","entity:Funnel"]` from the auto-tagger). `gates-engine.js:1741` compiles `gate.pattern` with `new RegExp()` against command text, so `entity:Customer+entity:Funnel` could never match — and `+` silently parsed as a quantifier.

Grouping by tag is legitimate. Reusing that key as the *match pattern* is not. Pattern now derives from the captured command via `contextToPattern()`, regex-escaped.

**2. The regression guard counted the incident as a false positive.** The command being learned was, by definition, **allowed before it was learned** — that prior allow is the failure the operator thumbs-downed. With `REGRESSION_FALSE_BLOCK_LIMIT = 0`, *every* gate learned from a real failure was quarantined to `warn`.

Incident contexts are now excluded from the false-block count by **normalized command equality**, not substring containment — so a different command that merely quotes the incident text still counts as genuine collateral damage. Wired through both the new-gate and the `warn`→`block` upgrade path; the upgrade path is the one the real flow takes.

Also: `promote()` now refuses to persist any gate whose pattern cannot match its own originating context.

## Evidence — clean sandbox, realistic path

```
STEP 1 — agent runs it, allowed (the incident):
    allow
STEP 2 — 3x 👎 -> promoted:
    action: block | occ: 3
STEP 3 — SAME command again:
    deny | [GATE:auto-entity-customer-entity-funnel]
STEP 4 — unrelated must still pass:
    allow   (git status)
    allow   (a dev-dependency install)
```

## Tests

`tests/auto-promote-gates.test.js` + `tests/auto-promote-regression-gate.test.js`: **51 pass / 0 fail**.

Five new regression tests, including one asserting a promoted gate denies the command that created it — the assertion whose absence let this ship.

Three existing tests keyed off `gate.pattern === '<tag>'`, which *encoded* defect 1; they now key off the stable `gate.id`. The quarantine test used the incident text itself as the "prior safe action," conflating incident replay with collateral damage; it now uses a genuinely different command that the pattern matches, preserving the real safety property.

**Pre-existing failure, not from this PR:** `run allows writes into private resume secrets vault` fails identically on clean `origin/main` when the 5 gate suites run together (cross-file env leakage). Baseline `origin/main` 296 pass/1 fail; this branch 301 pass/1 fail.

## Follow-up (not in this PR)

`public/index.html:1342` says "every 👎 compiles into a rule, and each rule…". That was not true before this change. Landing-page proof copy should be revisited once this lands.

Side note: creating this PR was initially blocked by our own `slopsquat-guard`, which read `(npm` in the body text as a typosquat of `pnpm`. Prose in a PR body is not a package install — worth a separate look at that gate's context sensitivity.

### DIFF
--- a/.changeset/auto-promoted-gates-enforce.md
+++ b/.changeset/auto-promoted-gates-enforce.md
@@ -1,0 +1,13 @@
+---
+"thumbgate": patch
+---
+
+Make auto-promoted gates actually enforce.
+
+The thumbs-down → block loop was inert. Three repeated 👎 on a command produced a gate that rendered as `action: "block"`, `severity: "critical"`, `occurrences: 3` — while a pre-action check on that exact command still returned **allow**. The dashboard showed a learned blocking rule; nothing was enforced.
+
+**The gate's match pattern was the group key.** `buildGateRule` reused `group.key` as the gate's `pattern`. Keys are usually tag-derived — a `kubectl delete deploy` capture was tagged `entity:Customer` / `entity:Funnel` by the auto-tagger. The gates engine compiles `pattern` with `new RegExp()` and tests it against command text, so `entity:Customer+entity:Funnel` could never match a command, and its `+` silently parsed as a regex quantifier. Grouping by tag is correct; reusing that key as the match pattern is not. Patterns now derive from the captured command, with regex metacharacters escaped.
+
+**The regression guard counted the incident itself as a false positive.** The command you are learning to block was, by definition, allowed before you learned it — that prior allow *is* the failure that got thumbs-downed. With a false-block limit of zero, every gate learned from a real failure was quarantined to `warn` and never enforced. Originating contexts are now excluded from the false-block count by normalized command equality, so genuine collateral damage (a different command that merely quotes the incident text) still quarantines as intended. Applied to both the new-gate and the `warn` → `block` upgrade path.
+
+`promote()` also now refuses to persist any gate whose pattern cannot match its own originating context, rather than writing an inert rule that reads as though the agent learned something.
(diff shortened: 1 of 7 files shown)


IgorGanapolsky:
Package boundary CI fail (`<= 401, got 403`) is from landing diagram assets on main, not the gate logic. Ratchet bump 401→405 is in #3120 (demo v2) — merge order: either land #3120 first or cherry-pick the package-boundary one-liner into this branch.

IgorGanapolsky:
Superseded by clean re-land + e2e + A+ demo: see the open PR on `fix/auto-promote-enforce-a-plus`. This branch is CONFLICTING with main — safe to close after the superseding PR merges.

IgorGanapolsky:
Superseded by #3125 (merged 8b3a515b) — same enforce fix + executable-action grouping + A+ demo e2e.

PR #3130, opened 2026-07-31
fix(packaging): declare YAML runtime dependency

IgorGanapolsky:
## Summary

- declare `js-yaml` as a production dependency because the packaged API server requires it directly
- add a package-boundary regression test that binds the runtime import to the production dependency declaration
- preserve the existing pinned version and override

## Verification

- `node --test tests/package-boundary.test.js tests/openapi-parity.test.js` (12/12)
- `npm ci --omit=dev`
- production-only `require('js-yaml')` and `require('./src/api/server.js')`
- pre-commit and pre-push guards

## Evidence boundary

The initial clean-suite run also observed the entire root `node_modules` directory missing before `test:api`. That disappearance has not reproduced: the API suite and every preceding stage now pass with dependencies intact. This PR fixes the independently reproducible packaged-runtime defect; it does not claim a cause for that one unexplained local anomaly.

### DIFF
--- a/.changeset/direct-js-yaml-runtime.md
+++ b/.changeset/direct-js-yaml-runtime.md
@@ -1,0 +1,5 @@
+---
+'thumbgate': patch
+---
+
+Declare the YAML parser used by the packaged API server as a production dependency so production-only installs can load the runtime without relying on a transitive development package.
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,7 @@
         "apache-arrow": "^18.1.0",
         "better-sqlite3": "^12.9.0",
         "dotenv": "^17.4.2",
+        "js-yaml": "4.3.0",
         "playwright-core": "^1.59.1",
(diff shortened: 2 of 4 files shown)


IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `273aa20fcb7fbacf9078259e937ba4afce1cb89a`
**Production buildSha:** `273aa20fcb7fbacf9078259e937ba4afce1cb89a`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","degraded":false,"version":"1.30.0","buildSha":"273aa20fcb7fbacf9078259e937ba4afce1cb89a","uptime":51.102064326,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

PR #3125, opened 2026-07-31
fix(gates): auto-promoted gates enforce + A+ demo learning loop

IgorGanapolsky:
## Summary
- **P0 fix:** auto-promoted gates match the **command**, not the tag group key (`entity:Customer+entity:Funnel` was inert). Unmatchable gates are not persisted. Originating incidents excluded from regression quarantine.
- **E2E test:** tagged 3× 👎 → promote → `gates-engine` **deny** on the originating command.
- **Demo A+:** `demo/scale-the-vibe-demo.sh --learn` proves `ALLOW → 3× 👎 → auto-promote → DENY` with command-derived pattern proof.

## Live proof
```text
✓ pattern is the command (not entity:Customer+entity:Funnel)
✓ A+ learning loop: ALLOW → 3× 👎 → auto-promote → DENY
```

## Supersedes
- #3119 (conflicted with main) — same fix re-landed cleanly + demo + e2e

## Test plan
- [x] `node --test tests/auto-promote-gates.test.js tests/auto-promote-regression-gate.test.js` (52/52)
- [x] `bash demo/scale-the-vibe-demo.sh --learn`

PR #3100, opened 2026-07-29
feat: prove and route local inference backends

IgorGanapolsky:
## Summary

- add a real OpenAI-compatible local inference proof command
- verify model discovery plus three deterministic chat completions
- measure contract pass rate, latency, token throughput, and routing readiness
- route to interactive, private batch, evaluation-only, or managed fallback lanes
- fail the proof command when the model is unavailable or violates the response contract
- support oMLX environment aliases without serializing API keys

## Verification

- `npm test` — passed
- `npm run test:api` — 969 passed
- `npm run test:high-roi` — 195 passed
- `node --test tests/local-model-profile.test.js` — 22 passed
- `npm run test:coverage` — 87.12% lines, 74.73% branches, 88.56% functions
- adapter proof, automation proof, self-healing check, version sync, and pre-push guards — passed
- `npm audit` — 0 vulnerabilities

## Live oMLX proof

`qwen3-0.6b-4bit` returned exact `READY` on 3/3 samples. p50 was 1,584 ms and p95 was 1,629 ms, producing `interactive_ready` / `interactive_local` with no blocked uses.

A patch changeset is included.

PR #3131, opened 2026-07-31
test(wiring): ratchet against capability modules with no caller

IgorGanapolsky:
## Why

Three components surfaced today that were well-written, fully tested, named after a capability — and wired to nothing:

| module | lines | state |
|---|---|---|
| `auto-promote-gates.js` | — | promoted gates carried tag-derived patterns that could never match a command. Fixed in #3125. |
| `judge-reward-function.js` | 408 | `buildCompositeReward` is called **only from its own test**; nothing injects a judge fn, so `scoringMode` is permanently `deterministic_only` |
| `cross-encoder-reranker.js` | 252 | named for a cross-encoder; actually an LLM reranker gated on `ANTHROPIC_API_KEY`, silently falling back to heuristics when the key is absent |

**Unit tests structurally cannot catch this.** Every one of those modules passes its own tests in isolation. The defect is the *absence of a caller*, which lives between modules, not inside one.

The first instance was expensive: the auto-promotion bug meant a promoted gate rendered in the dashboard as `action:block, severity:critical, occurrences:3` while enforcing nothing. It looked like the agent had learned. It hadn't.

## What this adds

`tests/capability-wiring-ratchet.test.js`, wired into the `test` chain as `test:capability-wiring`.

**Test 1 — orphan ratchet.** Pins the count of scripts whose basename appears nowhere in the tracked tree outside their own file and `tests/`. Baseline **19 of 475**, captured today. Same ratchet pattern already used by `tests/package-boundary.test.js`: the number may fall freely as dead code is wired or deleted; it cannot rise without a deliberate bump and a note.

**Test 2 — claim-bearing modules stay reachable.** A narrower, higher-signal assertion on `gates-engine`, `auto-promote-gates`, `lesson-retrieval` — the modules whose silence is most expensive, because the landing page and the pitch both assume they run.

## Design note

Matching is deliberately **permissive**: a `require`, an npm script, a workflow step, or even a docs mention all count as a reference. The goal is finding modules with *no reachable caller at all*, not enforcing import hygiene. False positives would make the check annoying and get it disabled; false negatives just mean it catches less. Given a ratchet only has to hold the line, that trade is correct.

Baseline is a count rather than an allowlist so the test does not need editing every time a script is legitimately added or removed.

## Verification

```
node --test tests/capability-wiring-ratchet.test.js tests/package-boundary.test.js
# pass 6
# fail 0
```

## Not in scope

Wiring `judge-reward-function.js` or routing `cross-encoder-reranker.js` off `ANTHROPIC_API_KEY` — the latter belongs to #3100 (`prove and route local inference backends`). This PR only makes the class visible so it stops recurring silently.


IgorGanapolsky:
All three findings fixed in `f81a1998`. Together they meant this test was close to a no-op — good catch.

**Exclude non-call references.** `package.json`'s `files[]` lists every packaged script by path, so the substring scan counted the packaging manifest as a call site and marked nearly the whole tree wired. Only the npm `scripts` block counts now: `node scripts/foo.js` is a real invocation, a packaging entry is not.

**Include nested scripts.** The scan read only immediate children of `scripts/`, leaving **69** modules under `scripts/lib`, `scripts/durability`, `scripts/social-analytics` and others permanently invisible. Now walks recursively.

**Ratchet against the real set.** The baseline was a guess. It is now the measured number.

**Corrected figures: 42 orphans across 544 scripts.** The original 19/475 was wrong on both terms.

One note against myself: I first wrote the guard-the-guard assertion too strictly — npm `scripts` legitimately contains `scripts/*.js` paths, so asserting none appear failed immediately. It now tests the property that actually matters: a module packaged but never invoked must not be rescued by its `files[]` listing.

4 pass / 0 fail.

IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `f7247219f89748019718264a733a961670c91b19`
**Production buildSha:** `f7247219f89748019718264a733a961670c91b19`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","degraded":false,"version":"1.30.0","buildSha":"f7247219f89748019718264a733a961670c91b19","uptime":35.086358997,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

PR #3133, opened 2026-07-31
chore(release): 1.31.0 via changeset:version — get the enforcement fix to npm

IgorGanapolsky:
## Why this release matters more than a version bump

`npm thumbgate@1.30.0` — gitHead `7a297924`, published **2026-07-29** — predates #3125. The published tarball still contains the original `extractPatternKey`:

```js
// Use tags as primary grouping key; fall back to context normalization
if (tags.length > 0) return tags.sort().join('+');
```

Verified directly against the published artifact:

```
extractExecutableAction : 0
contextToPattern        : 0
incidentContexts        : 0
```

That is the inert-gate bug. A user on 1.30.0 who thumbs-downs the same command three times gets a gate written to disk as `action: "block"`, `severity: "critical"`, `occurrences: 3` — with a match pattern like `entity:Customer+entity:Funnel` that can never match a command.

**The failure mode is the dangerous direction:** the dashboard shows a critical blocking rule, so it reads as "the agent learned." The next identical command runs anyway.

### The part that is easy to get wrong

The PreToolUse gate runs **on the user's machine, from the npm package** — not on the hosted service. `main` and the Railway deployment both carrying #3125 put the fix in *nobody's* hands. npm is the only distribution path that matters for enforcement, and it has been two days stale across the exact window where the self-improving claim went onto the landing page.

## What this PR does

Bumps `1.30.0 → 1.30.1` and propagates via `scripts/sync-version.js` (38 targets). **No functional change** — it ships what is already on `main`.

`publish-npm.yml` triggers on push to `main` touching `package.json`, so merging this is what actually releases.

## Pre-verified locally — the same checks the workflow gates on

```
sync-version.js --check     ✔ All 38 targets in sync at v1.30.1
test:deployment             132 pass / 0 fail
test:postinstall              4 pass / 0 fail
npm pack --dry-run          404 files, version 1.30.1   (ratchet ceiling 405)
```

And the check that actually matters — the packed artifact, not the green tick:

```
scripts/auto-promote-gates.js in the tarball:
  extractExecutableAction   : 5
  contextToPattern          : 4
  gateMatchesOwnContext     : 3
  incidentContexts          : 3
```

## After merge

Verify the published artifact rather than the workflow status — this repo has a documented history of "Publish to NPM" exiting green while shipping nothing:

```
npm view thumbgate version      # expect 1.30.1
npm view thumbgate gitHead      # expect the merge commit, not 7a297924
npm pack thumbgate@latest && grep -c extractExecutableAction package/scripts/auto-promote-gates.js
```


IgorGanapolsky:
Corrected. The first push hand-bumped 1.30.0 → 1.30.1 and the `Verify changeset` gate rejected it — correctly.

I had grepped the workflows for `changesets/action`, found none, and concluded version bumps were manual here. Wrong: the automation is an npm script (`changeset:version`), not a GitHub Action.

Reset to main and ran the real path. It consumed **19 pending changesets** and computed **1.31.0** — a minor, not the patch I hand-picked, because several changesets were `minor`. The hand-bump would have published under the wrong version.

That same mistake is the root cause of the staleness this PR fixes: 19 changesets accumulated because nobody ran `changeset:version`, and the gate only fires on PRs touching release-relevant files — so ordinary PRs merged green for days while the published version never moved.

Re-verified:
```
changeset:version         19 consumed, 0 remaining, CHANGELOG regenerated
sync-version.js --check   38/38 targets at 1.31.0
test:deployment           132 pass / 0 fail
test:postinstall            4 pass / 0 fail
npm pack --dry-run        404 files (ceiling 405), version 1.31.0
shipped auto-promote-gates.js carries extractExecutableAction, contextToPattern,
                                      gateMatchesOwnContext, incidentContexts
```

IgorGanapolsky:
Both threads addressed.

**P1 — adapter pins.** Fixed in `b12fa61d`. Root cause was `sync-version.js` listing `adapters/claude/.mcp.json` plus the `config.toml`/`opencode.json` siblings for claw, hermes and perplexity, but not those three adapters' own `.mcp.json`. Every bump left them on the prior release. Fixed the target list rather than the three files so the next bump cannot reintroduce it — coverage went 38 → 41 targets. Verified on this branch:
```
adapters/claw/.mcp.json -> thumbgate@1.31.1
```

**P2 — bare-feedback prevention rule.** This was a rule emitted by the auto-generated prevention-rules section, not hand-written policy, which is why the thread is marked outdated. That section has since regenerated; `AGENTS.md:169` now reads *"No active auto-generated prevention rules at this time."* The hand-written policy at `AGENTS.md:87` says the opposite of what was flagged — *"Reject vague signals (for example bare 'thumbs down') from memory promotion"* — so the standing instruction is correct and the offending entry is gone.

Worth flagging as a systemic point rather than a one-off: the generator briefly produced a rule that violated the very policy sitting 80 lines above it. That is the generator's vague-signal filter failing, and it will recur if a bare 👎 reaches promotion. `scoreFeedbackReward` in #3134 grades exactly this (bare corrections score 0.76 and fail the required rubric); wiring that into the generator's input filter is the durable fix, and belongs in its own change.

IgorGanapolsky:

PR #3138, opened 2026-07-31
feat(llm): OpenAI-compatible gateway provider — stop silently degrading without an Anthropic key

IgorGanapolsky:
Add an OpenAI-compatible gateway provider so LLM capabilities work without an Anthropic key.

Six runtime scripts gated on `ANTHROPIC_API_KEY` — `cross-encoder-reranker`, `eval-rag`, `self-distill-agent`, `secret-scanner`, `tool-registry`, `llm-client` — and every one degraded to heuristics **silently** when it was absent. `isAvailable()` was literally `Boolean(process.env.ANTHROPIC_API_KEY)`, so an operator paying for GLM or Kimi through a local gateway got the heuristic path with no indication the model tier was never consulted.

`llm-client` now falls back to any OpenAI-compatible endpoint (LiteLLM, vLLM, Ollama `/v1`, LM Studio) when no Anthropic client is available:

- `THUMBGATE_LLM_GATEWAY_URL` — enables the provider. Unset means no gateway; a published install never calls localhost because a port is open.
- `THUMBGATE_LLM_GATEWAY_MODEL` — defaults to `glm-5.2`.
- `THUMBGATE_LLM_GATEWAY_TOKEN` — optional; local gateways usually need none. Read at call time, never retained.

`describeInferenceAvailability()` reports **which** provider is live (`anthropic` / `gateway` / `none`, with a reason), so callers can report an honest scoring mode instead of inferring from a key's presence.

**Reasoning-model handling.** Models like GLM and Kimi split the response: `reasoning_content` holds the scratchpad, `content` holds the answer. Two empty-content cases that must not be conflated:

- `finish_reason: 'length'` — truncated mid-reasoning, the answer does not exist yet. Returns `null` so the caller falls back deterministically. Returning the scratchpad here would hand it confident-looking garbage.
- `finish_reason: 'stop'` — the model finished but emitted everything in the reasoning channel; that text is the answer.

Verified live against a LiteLLM gateway: `glm-5.2` returned `[0.9, 0.1]` for a relevance-scoring prompt (71 tokens, 23 cached), and the truncated variant correctly returned `null`.

No new dependency — the gateway contract is a single POST, and a security product's runtime needs a better reason than convenience to grow its supply chain.

## Provider audit on the operator machine

Probed the live LiteLLM gateway (31 models). Redacted provider errors — one leaked a key fragment.

| model | status |
|---|---|
| `glm-5.2`, `glm-coding`, `glm-turbo` | working |
| `kimi-code-k3` | working |
| `kimi-k3`, `kimi-coding` | **account suspended** |
| `hermes-local` | working (local) |
| `text-embedding` | Ollama `nomic-embed` missing |

Two findings for follow-up, not fixed here: the suspended Kimi routes, and the missing embedding model (which bears on retrieval quality).

## Tests

5 hermetic tests, no gateway required, wired as `test:llm-gateway`. Regression-clean: `llm-client` + `managed-lesson-agent` 19 pass / 0 fail.

Follow-up: `cross-encoder-reranker` still defaults `useLLM = false` and no caller overrides it, so its LLM path stays unreachable even with this landed. That is a separate change.


IgorGanapolsky:
All three findings addressed in `901a7cfb`. Each was correct.

**P1 — route self-distillation through the gateway.** `self-distill-agent.js:402` branched on `Boolean(process.env.ANTHROPIC_API_KEY)` directly, so it never reached the new fallback. Now calls `describeInferenceAvailability()`. An operator with a configured gateway was getting heuristic distillation while a working model sat one call away.

**P2 — forward supplied message histories.** This was a regression I introduced. `callGatewayInternal` built messages from `systemPrompt`/`userPrompt` only, so a caller passing `options.messages` had its whole history dropped and one empty user message sent, while the Anthropic path preserves it in `buildClaudeRequest`. Extracted `buildGatewayMessages()` so it is directly testable; it now forwards a supplied history verbatim.

**P2 — report the actual provider.** `managed-lesson-agent.js:84` converted `isAvailable()` into the hard-coded label `claude-haiku-4-5`, so a gateway-backed run persisted a manifest claiming Claude produced the lessons. That is a provenance lie in the exact artifact an operator would consult to audit those lessons. Now reports the model that actually ran.

The last two are the same defect class this PR exists to close — a capability reporting a provider it never consulted. Fixing `llm-client` alone was not enough while consumers tested for one vendor's key themselves; thanks for catching that.

2 new tests pin history-forwarding and honest model labelling. `llm-gateway-provider` + `llm-client` + `managed-lesson-agent` + `self-distill`: **26 pass / 0 fail**.

I have not resolved these threads — leaving that to the reviewer.

IgorGanapolsky:

PR #3220, opened 2026-08-04
fix(gates): union every promoted-gate store so lessons apply in all repos

IgorGanapolsky:
## Root cause of the outbound-send incident

The promoted-gate store resolved to exactly **one** location — repo-local when present, otherwise a per-project directory under home. A gate promoted from a failure in repo A was invisible in repo B.

Measured on one machine, same engine, same moment, only `cwd` differing:

```
cwd=ThumbGate -> 45 promoted gates
cwd=Resume    ->  4 promoted gates
```

The 41 missing gates included every one promoted from an outbound-send failure. An agent then sent outbound mail from the repository that could not see them. The gate that should have caught it showed `lastFiredAt: null` because it was never in scope there — not because the hook layer was dormant.

## Change

Union the global store, every per-project store, and any repo-local store, deduplicated by gate id with repo-local winning a collision.

```
ThumbGate 45 -> 92    Resume 4 -> 47    trading -> 52    unrelated dir -> 47
```

## Safety — measured, not assumed

| Check | Result |
|---|---|
| Gates newly visible in the affected repo | 43 |
| ...that match anything in a 51,356-line corpus | 3 |
| ...and all 3 are warn-level, not block | 0 new denials |
| Existing gate suites | 89/89 |
| Mutation: union returns nothing | 88/89 — one test catches it |
| New parity suite | 6/6 |

## A no-op I nearly landed

The first version of this fix changed nothing. `resolveFeedbackDir()` is itself cwd-dependent, so a "global" path derived from it resolved to the identical repo-local path and the union collapsed to a single entry. Re-running the original 45-vs-4 measurement is the only reason it surfaced.

`tests/global-promoted-gate-store.test.js` now asserts the two paths are distinct — the specific check that would have failed on that no-op.

## Scope

This restores gate reach, not teeth. Most of these gates carry prose patterns that cannot match a tool call and are warn-level, so they still would not have blocked the send. That is a separate fault, tracked on AGENT-259, and it is the one with teeth.

https://claude.ai/code/session_01NDsSTuZMjotPXfbpi6SNxe


IgorGanapolsky:
Closing stale/blocked PR to clear the board. Branch kept for recovery. Herdr-related work is concentrated in #3258; docs README path is #3262.

PR #3251, opened 2026-08-06
fix(gates+mcp): unblock the PR-thread livelock, measure blast radius in the right repo, remove 5 phantom tools

IgorGanapolsky:
## Three gate/MCP correctness defects, each mutation-proven

Found while trying to do ordinary work in this repo. All three were discovered by executing the control against known-bad input rather than reading it.

### 1. The PR-thread gate blocked its own escape hatch (livelock)

Once `pr_thread_resolution_verified_after_commit` was pending, the gate denied **every** tool call in the session — `Bash`, `Read`, `Write`, `Edit`, `ToolSearch`, `get_scope_state`, `gate_stats`, and `mcp__thumbgate__satisfy_gate` itself. `satisfy_gate` is the only tool that can clear it, so there was no reachable way out: an ordinary commit on a feature branch halted the session permanently and required a human to delete the state file.

The exemption list compared the raw `toolName` against bare names, but hook payloads deliver MCP tools as `mcp__<server>__<tool>`, so `mcp__thumbgate__satisfy_gate` never matched `satisfy_gate`. The gate's own comment already documented the intended behaviour — `isReadOnlyObservabilityTool` in the same file strips the prefix correctly; this predicate simply didn't.

Mutation proof (`tests/pr-thread-livelock.test.js`):

| assertion | without fix | with fix |
|---|---|---|
| MCP-prefixed `satisfy_gate` exempt | **FAIL** | pass |
| all evidence tools, both forms | **FAIL** | pass |
| gate still denies ordinary tools | pass | pass |
| bare name exempt | pass | pass |
| no leak to `mcp__evil__deploy_to_prod` | pass | pass |

The three guards pass both ways on purpose — they prove the fix didn't just disable the gate.

### 2. `workflow-sentinel` measured the wrong repository

`collectAffectedFiles()` never honored a `cd` in the command, so a 4-file commit inside a git worktree was scored against the *session* repo's unrelated dirty files — reported as "9 files across 6 surfaces", then "6 across 4" for a **single** file — and hard-blocked. Its own remediation ("split the change") couldn't help, because the number never described the change.

It also discarded `git add` pathspecs: `git add one-file.js` was measured as if it staged the whole tree.

Fixed via `resolveCommandRepoRoot()` and `gitAddPathspecs()`. `tests/workflow-sentinel-blast-radius.test.js` covers both, plus guards that `git add -A` / `git add .` still scan the whole tree and that an unresolvable `cd` falls back to the session repo rather than reporting zero files (which would silently disarm the gate).

### 3. Five MCP tools were advertised with no handler

`schedule`, `user_profile`, `session_search`, `webhook_deliver`, and `infer_lesson_from_history` shipped in `tools/list` with full input schemas while the dispatcher had no `case` for any of them — calling one hit `throw new Error("Unsupported tool: <name>")`.

`webhook_deliver` was the worst: it advertised sending messages to Slack, Teams, and Discord, so a client could plan an entire notification workflow around a capability that does not exist. Registry goes 98 → 93 tools.

`tests/mcp-tool-dispatch-parity.test.js` fails if any advertised tool lacks a dispatcher case, resolving the `get_reliability_rules` → `prevention_rules` alias first. Mutation-proven: 1 failing assertion naming all five before, 3 passing after.

## Verification

- `tests/pr-thread-livelock.test.js` — 5/5 (2 fail without the fix)
- `tests/workflow-sentinel-blast-radius.test.js` — 5/5
- `tests/mcp-tool-dispatch-parity.test.js` — 3/3 (1 fail without the fix)
- `tests/gates-engine.test.js` — 193 pass / 4 fail, **identical to the pre-change baseline** (verified by swapping in the original `workflow-sentinel.js` and re-running). Those 4 are pre-existing red on `main` — tests 71–74, the "warn by default / deny under strict" family — and are not addressed here.
- Pre-commit guards: package parity, version sync, gates + feedback tests all passed.

## Note on test wiring

`tests/workflow-sentinel-blast-radius.test.js` was initially reachable by **zero** npm scripts while `test-suite-parity` still reported green — CI would never have run it. Both new gate tests are now wired into `test:gates`. That parity guard looks weaker than its name suggests and probably deserves its own follow-up.

https://claude.ai/code/session_01YPQ3LLEkosBJrw8gx8u8XF


IgorGanapolsky:

PR #3252, opened 2026-08-06
fix(memory-guard): stop a corrupt entry from hard-blocking every action

IgorGanapolsky:
A truncated hook payload leaked into the store and was admitted as a recurring negative pattern. Its keywords were the payload's own field names - workspaceroot, workspace, thumbgate - which appear in every payload the agent sends. workspaceroot is long enough that isSpecificKeyword() treats it as decisive alone, so a SINGLE hit hard denied every Bash call in the repo, including the commands needed to diagnose it. Blocked 20 times, warned zero.

keywords() now drops envelope tokens, and pattern building rejects serialized payload fragments. Both describe the transport, never the error, so neither can be evidence of a recurring problem.

Mutation-proven: 4 of 5 assertions do not hold without the fix. The fifth - a genuine pattern still blocks the action it describes - holds both ways, so this does not disarm the guard. 52/52 across all memory-guard suites.

### DIFF
--- a/.changeset/memory-guard-envelope-tokens.md
+++ b/.changeset/memory-guard-envelope-tokens.md
@@ -1,0 +1,30 @@
+---
+'thumbgate': patch
+---
+
+Stop a corrupt feedback entry from hard-blocking every action in a repository.
+
+A truncated hook payload leaked into the feedback store and was admitted as a
+recurring negative pattern. Its keywords were the payload's own field names —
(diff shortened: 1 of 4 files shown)


IgorGanapolsky:
Re-triggering checks: required CI contexts (test, CodeQL, Verify changeset, Analyze JS) are missing from this PR rollup, so merge stays BLOCKED despite review count 0.

PR #3420, opened 2026-08-13
feat(mcp): model-carried session handles with principal binding

IgorGanapolsky:
## Summary

InfoWorld (2026-08-13): MCP didn't remove sessions — it handed correlation to the model. That is an identity/governance surface, which is ThumbGate's product.

Implements the article's high-ROI controls:

1. **Authorize every handle** against the authenticated principal + tenant on every request (no internal-tool exemption)
2. **Real entropy + short TTL** (144-bit id, HMAC fragment, default 15m, hard max 1h)
3. **Idempotency** on mint and on side-effect operations so a retry after a lost handle cannot create twin baskets/charges
4. **Multi-turn tests at depth 30** (`simulateMultiTurnFidelity`) including drop / swap / mutate / forge / cross-tenant failures

### How this helps us
- Wrong-account charge class (model carries valid-looking id for wrong tenant/principal) becomes a **deny**, not a silent success
- Model dropping a required handle mid-workflow fails closed (`MISSING_SESSION_HANDLE`)
- Hosted multi-tenant MCP: handle registry is the trust boundary the article says the spec left open

### API
- `scripts/mcp-session-handles.js` — `mintHandle`, `authorizeHandle`, `bindIdempotency`, `verifySessionHandle` (stub-compatible)
- Wired into `test:high-roi` + `test:mcp-session-handles`
- Bundle ceiling 452 → 453

## Test plan
- [x] `node --test tests/mcp-session-handles.test.js` — 23/23 pass
- [x] Multi-turn depth-30 faithful path + adversarial mutations
- [ ] CI on PR


IgorGanapolsky:
Addressed Codex P1/P2 review threads in `084187dbf`:

1. **Wire into MCP dispatch** — `adapters/mcp/server-stdio.js` `callTool` now runs `authorizeMcpToolCall` before the semantic firewall (disable via `THUMBGATE_DISABLE_MCP_SESSION_HANDLES=1`).
2. **No handle without principal** — `verifySessionHandle` returns `MISSING_SESSION_CONTEXT` when a carried handle has no context.
3. **Tenant always bound** — `authorizeHandle` always normalizes tenant (default `default`) and always compares; omitting tenant no longer skips cross-tenant checks.
4. **Idempotency claim-before-complete** — `claimIdempotency` → side effect → `completeIdempotency`; concurrent second claim returns `in_flight`.
5. **Load durable store on init** — `loadFromDisk()` at module require + `ensureDiskLoaded()` on paths.

Tests: 26/26 pass.

IgorGanapolsky:
## Fix pushed (`f0c5e9589`)

CI failure `feedback session tools support follow-up capture and finalization over MCP` was a **false positive collision**: feedback `sessionId` was treated as an MCP model-carried handle.

**Fix:** ambiguous arg keys (`sessionId`, `handle`, `basketId`, …) only match minted `mcp_h_*.*` tokens. Explicit keys (`sessionHandle`, `handleId`, …) unchanged.

Local: `tests/mcp-session-handles.test.js` + `tests/mcp-server.test.js` → 75 pass / 0 fail. Unresolved review threads: 0.

PR #3398, opened 2026-08-12
feat(receipts): canonical request digest & HMAC signed execution receipts

IgorGanapolsky:
## Summary
- **Broker-Signed Execution Receipts**: Added `computeCanonicalRequestDigest`, `signReceiptDigest`, and `verifyReceiptSignature` to `scripts/action-receipts.js`.
- **Receipt Payload Enrichment**: `normalizeReceipt` now captures `principal`, `target`, `decision`, `idempotencyKey`, `providerEventId`, `requestDigest`, and `signature`.
- **Cryptographic Verification**: Added HMAC-SHA256 signature verification over canonical request digests.
- **Unit Tests**: Added tests to `tests/action-receipts.test.js` (8/8 PASSED).

## Verification Evidence
- `node --test tests/action-receipts.test.js`: 8/8 tests PASSED.
- Changeset: `.changeset/broker-signed-action-receipts.md`

PR #3421, opened 2026-08-13
feat(mcp): verify model-carried session handles before tool execution

IgorGanapolsky:
## Why

[MCP didn't remove sessions, it handed them to the model](https://www.infoworld.com/article/4208733/mcp-didnt-remove-sessions-it-handed-them-to-the-model.html) — the Aug-2026 stateless shift moved session correlation out of the protocol and into model-carried arguments.

The dangerous failure is not a malformed handle. It is a **well-formed, correctly signed, correctly-tenanted handle pointing at the wrong resource**, because the model lost the thread across turns. The server sees a legitimate authorized call and nothing logs an error. The article cites a Red Hat incident where an agent hit the wrong customer record for a four-figure sum, and ICLR 2026 work showing a 39% average performance drop in multi-turn settings.

**This is the exact class ThumbGate exists to catch, and we did not catch it.**

## What was there before

An untracked, unwired `scripts/mcp-session-handle-verifier.js` carrying the same defect class as #3398's P1 findings:

- `verifySessionHandle` split the handle, read `tenantId` out of it, and **never recomputed the HMAC** — the signature was decorative, so `mcp_sess_victim_aaaa.deadbeef` passed
- `expiresAt` was minted but never read; the 1-hour TTL was fiction
- fell back to a hardcoded signing value, publicly known and therefore forgeable
- no resource binding at all, so the article's core case sailed straight through

## What this does

Four checks, ordered by how often they bite:

1. **Signature** recomputed, constant-time compared
2. **Expiry** actually read
3. **Tenant** taken from the *verified* payload, never parsed from the untrusted string first
4. **Resource binding** — catches a genuine, unexpired, same-tenant handle aimed at a resource it was not minted for

Canonical payload is length-prefixed so no field can impersonate a delimiter boundary. Minting and verifying now throw unless signing material is configured; no shipped default.

## Tests — 14/14 green

Including the article's own recommendation to *test at turn depth thirty rather than turn one*: a 30-turn simulation where an agent holds two handles and reaches for the wrong one at turn 30. Asserts **exactly one denial, on turn 30, and none before**.

Backward compatible: handles minted without a resource binding do not block resource-scoped calls.

Registered in the npm test chain; `test-suite-parity` 5/5 green.

https://claude.ai/code/session_01Jp3JxyFffuyLLZHTy4XA8Y


IgorGanapolsky:
Superseded by #3420 (`feat/mcp-session-handles`), which merged to main at `75a7e2336eed` with principal-bound model-carried MCP session handles, idempotency, multi-turn fidelity tests, and MCP dispatch wiring (including the domain-`sessionId` collision fix).

Closing this sibling verification-only PR to avoid double-shipping the same surface.

PR #3630, opened 2026-08-24
fix(tests): make gates-engine suite hermetic against inherited strict enforcement

IgorGanapolsky:
### Summary
Five tests in `tests/gates-engine.test.js` assert the default warn-by-default posture but read the enforcement mode from the ambient process environment. When `THUMBGATE_STRICT_ENFORCEMENT=1` is exported by the operator's shell or an agent harness, strict mode turns every warn into a hard deny and replaces `additionalContext` with `permissionDecisionReason`, causing false CI failures on a clean tree.

### Changes
- Capture `THUMBGATE_STRICT_ENFORCEMENT` into `ORIGINAL_ENV` block
- Delete in `beforeEach` and restore in `afterEach`
- Add changeset

### Verification
- `node --test tests/gates-engine.test.js` passes 100% with and without strict mode set in ambient env

### DIFF
--- a/.changeset/gates-engine-strict-hermetic.md
+++ b/.changeset/gates-engine-strict-hermetic.md
@@ -1,0 +1,11 @@
+---
+'thumbgate': patch
+---
+
+fix(tests): make `tests/gates-engine.test.js` hermetic against an inherited `THUMBGATE_STRICT_ENFORCEMENT`
+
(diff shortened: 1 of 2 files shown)


IgorGanapolsky:
Superseded by #3631 (`cd19a7db`). This branch is the hermetic `THUMBGATE_STRICT_ENFORCEMENT` isolation only, so required \`test\` fails the same CodeQL lockstep pin that redded \`main\` after #3622. #3631 is the same unique commit plus the v4.37.7 \`ff2f1c62\` pin.

Closing as duplicate. No approval. No admin.

IgorGanapolsky:
Duplicate of #3631 (same hermetic test plus CodeQL lockstep pin).

PR #3673, opened 2026-08-25
feat(metrics): measure what the user waited, not what the server was busy for

IgorGanapolsky:
Stolen from NVIDIA's SemiAnalysis **AgentX** methodology. The GPU hardware in that post is irrelevant to us; the measurement discipline is not.

## The idea worth taking

AgentX reports two rates **side by side**, and treats the gap between them as the finding:

| Metric | Definition (verbatim) |
|---|---|
| Standard Interactivity | *"output tokens divided by the elapsed time from the first token to the last"* |
| **E2E Normalized Interactivity** | *"average output-token rate per user over the full request, calculated by dividing total output tokens by the time from request submission to final-token delivery"* |

Standard starts the clock at the **first token**. Queue time, scheduling, a runner that hasn't claimed the work — all invisible. E2E starts at **submission**, so it counts what the person counts.

## This is not theoretical — it is today's incident

**2026-08-25 production:** task created `17:34:08Z`, completed `17:34:53Z`. Execution was quick, the UI advertised a **"90s lease"**, and every server-side measure said healthy. The user spent 45 seconds watching a spinner reading *"Waiting for the fenced VPS runner to pick this up."*

```
Standard Interactivity:  ~100 tok/s   ← "this request was fast"
E2E Normalized:          ~11 tok/s    ← what the person experienced
```

**A metric that cannot see queue time cannot see that outage.** The test suite pins that ~9x divergence using the incident's real timestamps, so this class of defect surfaces as a number instead of a screenshot.

## Session-level, because agentic traffic isn't request-shaped

Also AgentX: *"Agentic sessions are long, stateful, and variable: they chain model calls, tool use, and growing context rather than following a fixed prompt-and-response pattern."* Their sample session grows ~60K → ~400K tokens; they cite a single agentic request at ~15x the tokens of ordinary chat.

So this aggregates whole sessions and **keeps tool gaps in the denominator**. A tool call the user waits through is wait, whoever executes it — a serving stack doesn't get to subtract offloaded time from its own score.

Also included:

- **`contextGrowth`** — first/last prompt tokens and growth factor. A fixed 8K-in/1K-out benchmark would call a 60K→400K session the same workload.
- **`compareRuns`** — refuses to compare two runs unless they replayed **the same sessions in the same order**. AgentX: *"Because every system receives the same recorded traffic, observed differences reflect the serving stack rather than benchmark-specific tuning."* Comparing over different traffic measures the traffic. That refusal is a tested invariant, not a warning.

## Fails closed

- **One unmeasurable turn makes the whole session `UNKNOWN`, never fast.** Measuring only the good turns would understate the wait — the original failure mode.
- **A zero-length generation window yields `null`, not `Infinity`.** An unmeasurable rate is not an infinitely fast one, and `Infinity` would flatter exactly the pathological case this watches for.
- Malformed or out-of-order timestamps are named in the output, not swallowed.

## Tests

14 assertions, all passing. **36/36** including `test-suite-parity`, `package-boundary`, `public-bundle-ratchet`, `public-core-boundary`, `pack-runtime-integrity`. The test target is registered in `package.json` — I learned this session that `test-suite-parity` enforces that, and deferring it only moves the failure later.

## What this does not do

Not wired into the dashboard or the task API yet. This lands the metric and its contract so the numbers can be reviewed on their own; instrumenting `submittedAt`/`firstTokenAt` at the call sites is the follow-up, and it's where the real work is.

## Merge-order note

Bumps the three file-count ceilings 512 → 513 and the size ceiling 8.01 → 8.03 MB (main sits at **99.9%** of the old size limit, so any addition trips it). **#3667 makes the same +1 claim** — whichever lands second needs a one-line rebase. That shared-integer ceiling is a coordination hazard worth its own fix.


IgorGanapolsky:
### Hygiene (2026-09-04)

Status: `DIRTY` / conflicting with `main` tip `ab232243`.

Not closing (<7d / still has unique work). Needs rebase onto current `main` in an isolated worktree, then re-run required checks (`test`, CodeQL, Verify changeset, Socket, GitGuardian, Analyze JS).

Open P1 Codex threads (if any) must be fixed in code before `resolveReviewThread` — conversation-resolution is required on `main`.

IgorGanapolsky:
Hygiene 2026-09-04: `ci=SUCCESS` but `mergeStateStatus=DIRTY` / CONFLICTING vs tip `b607eeb`. Needs rebase (or successor branch) before Trunk. Not closing (<7d / unique work).

IgorGanapolsky:
### Vercel check on this PR is stale / non-required

Production is Railway. The Vercel GitHub link for `thumbgate` is **disconnected** (2026-09-06), and tip already has `vercel.json` `git.deploymentEnabled: false` + ignore step `exit 0`.

If you still see a Vercel row (including `upgradeToPro=build-rate-limit`), it is a **historical** status from before disconnect — not a live deploy failure and **not** a required merge gate (`config/merge-quality-checks.json` optionalChecks).

See `docs/agents/vercel-hobby-not-prod.md`.

IgorGanapolsky:
Board-loop: **DIRTY/CONFLICTING** (ci=SUCCESS mss=DIRTY). Needs successor rebase onto `origin/main`, not `--admin`. Not stale-closed (<30d failing PRs stay).

Issue #3693, opened 2026-08-26
Agentic Design Patterns (Gulli) gap map: routing before lexical gates, executable remediations, per-gate FP telemetry, scope isolation


IgorGanapolsky:
## Map Gulli's Agentic Design Patterns taxonomy to ThumbGate gaps (CEO directive 2026-08-26)

Source: https://github.com/evoiz/Agentic-Design-Patterns (Antonio Gulli, 21 chapters + notebooks, 3k stars). ThumbGate already IS Ch-18 Guardrails and shipped Ch-16 Resource-Aware today (performance budgets, #3683). The measured gaps, grounded in the 2026-08-26 session's evidence:

1. **Ch 2 Routing — semantic gate classification.** The session logged 25+ documented lexical false-fires (a bare commerce word inside a git flag, a finance word inside a read-only audit URL, memory-guard locking onto repeated command shapes, filenames triggering hook self-protection — and this very issue's first submission was denied because its example text sat near a verb). The routing pattern = classify intent before dispatching to a rule. Proposal: a routing layer in gates-engine that classifies tool-call intent (read/diagnose vs mutate vs transact) BEFORE lexical rules run, so read-only diagnosis never reaches transaction rules. Complements #3682 (sequence/chain detection).
2. **Ch 12 Exception Handling & Recovery — every deny must carry one verified-executable remediation.** Several gates advertise remediations that are themselves broken or blocked (retrieve_lessons schema failure is cited as the first remediation on every workflow-sentinel block). Add a CI test that executes each advertised remediation path.
3. **Ch 19 Evaluation & Monitoring — per-gate precision telemetry.** Deny receipts exist (denies.jsonl); missing is per-rule false-positive rate on the dashboard (deny followed by a sanctioned workaround = candidate FP). The alert-noise ledger (#3667, merged today) provides the session plumbing.
4. **Ch 8 Memory Management — shared scope-store isolation.** Two sessions clobbered each other's task scope twice on 2026-08-26. Book pattern: per-agent state isolation; existing lesson proposes pinning scope per MCP process — this issue tracks implementing it.

Notebooks worth mining for test fixtures: Chapter_18 (guardrails), Chapter_12 (exception handling), Appendix_G (coding agents).

IgorGanapolsky:
## Triage (2026-09-03)

**Status:** OPEN — umbrella gap map.

[...]

Issue #3684, opened 2026-08-26
Reduce guard false positives in documentation and SQL


IgorGanapolsky:
The financial mutation guard rejects inert documentation and SQL when generic action words and business terms appear anywhere in the same payload. Duplicate project-local hook registration also doubles evaluation latency.

Acceptance:
- File-content tools permit inert currency, invoice, and SQL examples.
- Non-file actions require an action/object distance of 80 characters or less.
- Vendor prose is allowed while interactive checkout actions remain denied.
- Direct purchase and guard-tampering controls remain denied.
- Primary and hardened runtime copies remain byte-identical and locked.
- Regression tests cover allowed and denied cases.

PR #3697, opened 2026-08-27
fix(gates): left word boundary for the self-protect process rule

IgorGanapolsky:
## Problem

The self-protect process rule in the default gate policy-file has no left anchor, so any word that merely **ends** in the protected verb, followed by a space and (anywhere later) a protected process name, hard-blocks a harmless command. Observed 2026-08-26: a plain echo about a newly registered agent capability was denied five separate times across two sessions (receipts in the local gate history; it also bit the commit message of this very fix).

Related umbrella issues: #3684 (guard false positives), #3693 (gap map: routing before lexical gates).

## Fix

One-line pattern change — prepend a left-boundary group (start-of-string or non-word char) before the verb alternation:

- benign prose (word-suffix + later process name) no longer matches
- genuine terminations still deny: start-of-string (the existing hard-floor case) and mid-string after a non-word char are both covered

## Verification

- tests/gates-engine.test.js: **204 pass / 0 fail**, including a new regression test reproducing the 2026-08-26 false positive and asserting the real-invocation case still denies
- tests/gate-golden-set.test.js: 0 fail
- Standalone regex proof: 4/4 behavior checks (benign sentence, p-variant, bare verb, word-suffix)

Changeset included (patch).

https://claude.ai/code/session_01L2iHT9zUBjeDkrPUK5yS2a

### DIFF
--- a/.changeset/gate-word-boundary-self-protect.md
+++ b/.changeset/gate-word-boundary-self-protect.md
@@ -1,0 +1,8 @@
+---
+"thumbgate": patch
+---
+
+self-protect process rule now requires a left word boundary, so prose containing
+words that merely end in the verb (e.g. "skill " followed later by a protected
+process name) no longer hard-blocks harmless commands; real kill/pkill/killall
+invocations still deny. Regression test added from the 2026-08-26 false positive.
--- a/config/gates/default.json
+++ b/config/gates/default.json
@@ -509,7 +509,7 @@
       "toolNames": [
         "Bash"
       ],
-      "pattern": "(?:kill|pkill|killall)\\s+.*(?:thumbgate|gates-engine|budget-enforcer)",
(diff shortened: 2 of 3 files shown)


IgorGanapolsky:
## ✅ Deploy verified

**Merge commit:** `e7afff0300c08c765e2eabca14c24df2537a90b6`
**Production buildSha:** `e7afff0300c08c765e2eabca14c24df2537a90b6`
**buildSha match:** true

**/health snapshot:**
```json
{"status":"ok","degraded":false,"version":"1.35.0","buildSha":"e7afff0300c08c765e2eabca14c24df2537a90b6","uptime":67.11002937,"checks":{"feedbackDir":{"ok":true},"hostedConfig":{"ok":true},"buildMetadata":{"ok":true}},"deployment":{"appOrigin":"https://thumbgate.ai","billingApiBaseUrl":"https://thumbgate.ai"}}}
```

- Fail: `none`

